PR #35, opened 2022-12-01
Docs: Setup i18n  & Add spanish translation for `docs/index.md`

rafnixg:
- Setup multi-language docs
- Translate index.md (Overview)
- Translate Nav menu


mkdocs.yml
```diff
@@ -41,6 +41,37 @@ plugins:
           rendering:
             show_root_heading: true
             show_source: false
+  - i18n:
+      default_language: en
+      languages:
+        en:
+          name: English
+          build: true
```
jowilf, line 49: I prefer `build=false` here to avoid duplication
```suggestion
          build: false
```

docs/index.es.md
```diff
@@ -0,0 +1,112 @@
...
+- Exportación de datos a CSV/EXCEL/PDF e impresión con navegador
+- Autenticación
+- Autorización
+- Administrar archivos
+- Vistas personalizadas
+- ORMs compatibles
+    * [SQLAlchemy](https://www.sqlalchemy.org/)
+    * [SQLModel](https://sqlmodel.tiangolo.com/)
+    * [MongoEngine](http://mongoengine.org/)
+    * [ODMantic](http://mongoengine.org/)
```
jowilf, line 46: This link is not the good one. I will update for english version
```suggestion
    * [ODMantic](hhttps://github.com/art049/odmantic/)
```

jowilf (changes requested):
Really thank you for this work, you are doing great.

For more readability in long term (when maybe we will have a lot of languages), i think we should move each page inside a directory. Otherwise we should use folder based docs structure . But  I prefer suffix based docs structure with each page inside a directory

So, the new location for `index.md` and `index.es.md` should be `docs/index/index.md` and `docs/index/index.es.md`

rafnixg:
> Really thank you for this work, you are doing great.
> 
> For more readability in long term (when maybe we will have a lot of languages), i think we should move each page inside a directory. Otherwise we should use folder based docs structure . But I prefer suffix based docs structure with each page inside a directory
> 
> So, the new location for `index.md` and `index.es.md` should be `docs/index/index.md` and `docs/index/index.es.md`

I already made the change to the new folder structure, the file index.md and index.es.md are the only ones that are outside a folder, so that Mkdocs recognizes the entry point

jowilf (approved):

jowilf:
Thank you for your contribution

rafnixg:
in the next few days I will translate what is missing, thanks to you for this library!

Issue #21, opened 2022-11-10
feat: actions feature


giaptx:
The first, thanks for your great package. :+1: 

I'm trying to use this package to create my admin dashboard, but I can not see the actions feature.
For example https://docs.djangoproject.com/en/4.1/ref/contrib/admin/actions/

Could you pls point me how to do it? And do you have any plan for that

Thank you very much!

jowilf:
Hello @giaptx,
I'm sorry but this feature is not currently available,
But I plan to add this very soon,
Stay tuned

rafnixg:
Hello, thanks for this package.
I have a question on this topic, is there a way to implement this?
@giaptx  how could you solve this?

jowilf:
Hello @rafnixg,

Actually, when you select items you have only one possibility (DELETE), This is already a bulk action. Just need to replace the delete button by dropdown and accept custom actions exactly like flask-admin (https://flask-admin.readthedocs.io/en/latest/_modules/flask_admin/actions/)

rafnixg:
Hello @jowilf 
I have tried to do what you explain but I still can't do the custom actions creating the decorator, have you made some progress on this feature or do you have some code that can guide me, thank you very much!

jowilf:
Hello @rafnixg,
i will work on this just after 0.3.0 release. 
Sorry, i can't do more right now

giaptx:
@rafnixg I'm just waiting, 

Thank you, @jowilf. Do you need any help? I'm not a master on frontend stuff.

jowilf:
Hello @giaptx,
I'm doing by best to make it available by the end of this week,

But yes any help are welcomed, even if you are not good on frontend stuff, you can still help for the doc :)
Thank you

jowilf:
Hello @giaptx,
I'm doing by best to make it available by the end of this week,

But yes any help are welcomed, even if you are not good on frontend stuff, you can still help for the doc :)
Thank you

rafnixg:
@jowilf I would like to be able to contribute to the documentation, my main language is Spanish, I could start with the translation and reach more people

jowilf:
Thank you @rafnixg,
Let me know if you need some action from me for multi-language support

rafnixg:
@jowilf don't worrie in my fork, i have setup the multilingual featurem using the mkdocs-static-i18n lib.

soon I do pull request with the initial configuration and the first page translated to start.

giaptx:
[...]

PR #61, opened 2022-12-17
Allow custom form for batch actions

giaptx:
Hi @jowilf,

I would like to create a PR to try to resolve this issue https://github.com/jowilf/starlette-admin/issues/21#issuecomment-1354299135

I'm not master on frontend, but I try to make this works for my application. The idea is each action will have own form html to custom the input for each action.

Could you have a look? Please feel free to reject if you have any better plan or solution.

Thanks


starlette_admin/actions.py
```diff
@@ -7,6 +7,7 @@ def action(
     confirmation: Optional[str] = None,
     submit_btn_text: Optional[str] = "Proceed",
     submit_btn_class: Optional[str] = "btn-primary",
+    form: str = "",
```
jowilf, line 10:
```suggestion
    form: Optional[str] = None,
```

starlette_admin/statics/js/list.js
```diff
@@ -423,6 +423,11 @@ $(function () {
       query.append("pks", s);
     });
     query.append("name", name);
+
+    $('#modal-action :input').each(function() {
+        query.append(this.name, $(this).val());
+    });
+
```
jowilf, line 430: Send form data as [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData)

jowilf: Useful link -> https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects#retrieving_a_formdata_object_from_an_html_form

examples/custom_actions/models.py
```diff
@@ -36,14 +36,20 @@ class ArticleView(ModelView):
         confirmation="Are you sure you want to mark selected articles as published ?",
         submit_btn_text="Yes, proceed",
         submit_btn_class="btn-success",
+        form="""<div class="form-floating mb-3">
```
jowilf, line 39: I'm not sure but i think a template path could be better here, wdyt?

jowilf (changes requested):
Hello @giaptx, 
It's better idea, we should find a better and easy way to include that,
I made some few comments, please check them

Please add some test, update docstring and the documentation

giaptx:
Thanks for your review. I will improve when I get time.
Feel free to push to PR if you get time too, because you are best.

jowilf (approved):

jowilf:
Thank you all for your contributions

PR #68, opened 2022-12-27
Fix searchFormat not working for `DateField` & `TimeField`

ihuro:
Copy the `model.fields` list before apply the `shift` function to keep original values un the `model` object. Otherwise, other functions cannot find fields attributes. For example, the extraction of the `search_format` in https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/statics/js/list.js#L164 does not work.

### DIFF
--- a/starlette_admin/statics/js/utils.js
+++ b/starlette_admin/statics/js/utils.js
@@ -52,3 +52,10 @@ function pretty_print_json(data) {
     .replace(/>/g, "&gt;")
     .replace(jsonLine, replacer);
 }
+
+if (typeof structuredClone === 'undefined') {
+  // Simple (non-performant) replacemente of `structuredClone` for old browsers
+  structuredClone = function (value) {
+    return JSON.parse(JSON.stringify(value));
+  }
+}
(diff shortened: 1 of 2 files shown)


jowilf:
Hello @ihuro, 
Thank you for the catch but your solution doesn't deep copy nested fields

ihuro:
> Hello @ihuro, Thank you for the catch but your solution doesn't deep copy nested fields

I've updated the PR with a new proposal.
The function `structuredClone` is supported in newer browsers versions https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility

jowilf:
Thank you for your contribution

starlette_admin/statics/js/utils.js
```diff
@@ -52,3 +52,10 @@ function pretty_print_json(data) {
     .replace(/>/g, "&gt;")
     .replace(jsonLine, replacer);
 }
+
+if (typeof structuredClone === 'undefined') {
+  // Simple (non-performant) replacemente of `structuredClone` for old browsers
```
jowilf, line 57:
```suggestion
  // Simple (non-performant) replacement of `structuredClone` for old browsers
```

jowilf (changes requested):

PR #109, opened 2023-03-01
missed "async" fixed

Ilya-Green:
Without it leads to error when trying to submit wrong login and password

### DIFF
--- a/examples/auth/provider.py
+++ b/examples/auth/provider.py
@@ -56,7 +56,7 @@ class MyAuthProvider(AuthProvider):
 
         return False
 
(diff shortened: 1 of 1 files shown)


Ilya-Green:
it seems that the same must be fixed inside the library in "MyAuthProvider" class

jowilf:
This method is called inside the template, and can not be async. If you have a use-case who required async get_admin_user, Please fill a [feature request issue template](https://github.com/jowilf/starlette-admin/issues/new?assignees=&labels=enhancement&template=feature_request.md&title=Enhancement%3A+)

Issue #182, opened 2023-05-04
Bug: 303 redirects at login screen


mixartemev:
**Describe the bug**
Unable to see logo (from statics dir) at login screen, and also 1 of scripts (`js.cookie.min.js`) is not in allowed paths in `AuthMiddleware`.
<img width="585" alt="Screenshot 2566-05-04 at 16 48 33" src="https://user-images.githubusercontent.com/5181924/236230563-a9300b22-4375-41e9-9b7c-6b940acd36d0.png">

**To Reproduce**
Set `statics_dir` and put login_logo in this dir.
<img width="565" alt="Screenshot 2566-05-04 at 16 56 05" src="https://user-images.githubusercontent.com/5181924/236230638-56254751-a01a-43c2-ad01-76746367ee79.png">

**Environment (please complete the following information):**
 - Starlette-Admin version: 0.8.1
 - ORM/ODMs: SQLAlchemy

jowilf:
Fix in #187 & #183

PR #183, opened 2023-05-04
Allow Unauthenticated Access to js.cookie.min.js

mixartemev:
#182

### DIFF
--- a/starlette_admin/base.py
+++ b/starlette_admin/base.py
@@ -133,7 +133,11 @@ class BaseAdmin:
                 [] if self.middlewares is None else list(self.middlewares)
             )
             self.middlewares.append(
-                Middleware(AuthMiddleware, provider=self.auth_provider)
+                Middleware(
+                    AuthMiddleware,
+                    provider=self.auth_provider,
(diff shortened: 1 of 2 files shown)


starlette_admin/base.py
```diff
@@ -133,7 +133,11 @@ def init_auth(self) -> None:
                 [] if self.middlewares is None else list(self.middlewares)
             )
             self.middlewares.append(
-                Middleware(AuthMiddleware, provider=self.auth_provider)
+                Middleware(
+                    AuthMiddleware,
+                    provider=self.auth_provider,
+                    allow_paths=[self.login_logo_url],
+                )
```
jowilf, line 140:
```suggestion
                Middleware(AuthMiddleware, provider=self.auth_provider)
```
You can now allow additional paths from your AuthProvider by using the allow_paths parameter (#187). This makes it much easier. You can check out an example implementation at:  https://github.com/jowilf/starlette-admin/blob/6e07f49d11643a7bfca1fb391990500cc3ee095f/examples/auth/app.py#L46

jowilf:
Thanks for your contribution 🎉

PR #221, opened 2023-07-01
Enhance AuthProvider to support oauth2/OIDC flow

### LINKED ISSUE #218: Enhancement: Add option to authenticate via oauth2/OIDC instead of username/password
**Is your feature request related to a problem? Please describe.**
As far as I understood username & password is currently the only option to implement admin authentication. E.g. the default login screen is always rendered even if you change the login_path manually when instantiating the admin-backend

**Describe the solution you'd like**
It would be nice to use an oauth2 flow to authenticate users.
So far I used an authlib-oauthclient to handle the oauth2flow.

With my oauthclient I need to call 2 methods to complete the sign-in:
- authlib_client.authorize_redirect(<redirect_uri>) -> This sends the user to the OAuth-provider for authentication, the user is then sent back to our application to the specified redirect_uri
- authlib_client.authorize_access_token() -> coming back from the oauthprovider the OAuth client validates the token/userdata. If successful some data is stored in the current session (e.g. "isAdmin"=True if the user has admin role in his userdata)

After the login steps the standard "is_authenticated" method is used to validate if and what the user is allowed to see

What is needed to facilitate the OAuth flow:
- if OAuth authentication is used, you should be able to run a custom login function (which sends the user to the external login_url of your OAuth-provider without showing the login screen/template)
- [optional] You need to be able to handle the response and store data in the session -> optional because this could be done with a separate fastapi route

**Describe alternatives you've considered**
You could implement the whole login flow with 2 separate fastapi-endpoints completely outside of the starlette-admin context (login & response validation) -> then you just need the option to disable the login screen and

**Additional context**

jowilf:
close #218

### DIFF
--- a/examples/authlib/provider.py
+++ b/examples/authlib/provider.py
@@ -1,0 +1,80 @@
+from typing import Optional
+
+from starlette.datastructures import URL
+from starlette.middleware import Middleware
+from starlette.requests import Request
+from starlette.responses import RedirectResponse, Response
+from starlette.routing import Route
+from starlette_admin import BaseAdmin
+from starlette_admin.auth import AdminUser, AuthMiddleware, AuthProvider
+
+from authlib.integrations.starlette_client import OAuth
+
+from .config import AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_DOMAIN
+
+oauth = OAuth()
+oauth.register(
+    "auth0",
+    client_id=AUTH0_CLIENT_ID,
+    client_secret=AUTH0_CLIENT_SECRET,
+    client_kwargs={
+        "scope": "openid profile email",
+    },
+    server_metadata_url=f"https://{AUTH0_DOMAIN}/.well-known/openid-configuration",
+)
+
+
+class MyAuthProvider(AuthProvider):
+    async def is_authenticated(self, request: Request) -> bool:
+        if request.session.get("user", None) is not None:
+            request.state.user = request.session.get("user")
+            return True
+        return False
+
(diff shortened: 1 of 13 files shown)


examples/authlib/provider.py
```diff
@@ -0,0 +1,80 @@
...
+            photo_url=user["picture"],
+        )
+
+    async def render_login(self, request: Request, admin: BaseAdmin):
+        """Override the default login behavior to implement custom logic."""
+        auth0 = oauth.create_client("auth0")
+        redirect_uri = request.url_for(
+            admin.route_name + ":authorize_auth0"
+        ).include_query_params(next=request.query_params.get("next"))
+        return await auth0.authorize_redirect(request, redirect_uri)
```
simonsax, line 47: I had to wrap redirect_uri in str() to avoid an error message -> return await auth0.authorize_redirect(request, str(redirect_uri))

examples/authlib/provider.py
```diff
@@ -0,0 +1,80 @@
...
+
+    async def render_logout(self, request: Request, admin: BaseAdmin) -> Response:
+        """Override the default logout to implement custom logic"""
+        request.session.clear()
+        return RedirectResponse(
+            url=URL(f"https://{AUTH0_DOMAIN}/v2/logout").include_query_params(
```
simonsax, line 53: this logout domain is provider specific -> my oauth2 provider uses a different endpoint <domain>/logout.do

simonsax (changes requested):
I tried it with my project and was able to implement oauth2 authentication. Thanks a lot!

sternalon:
Looking forward to this one being merged into the main branch!

PR #302, opened 2023-09-20
🎬 Detail actions

### LINKED ISSUE #301: Enhancement: Custom actions on detail page
**Is your feature request related to a problem? Please describe.**
Custom actions are only available to trigger on the list page.

**Describe the solution you'd like**
Custom actions should appear as buttons on the detail page. Actions should be able to be excluded from the detail page with a custom parameter.

**Describe alternatives you've considered**
Creating custom buttons currently requires overwriting the entire detail page for each model, for each action.

**Additional context**
I have a working example, and will be submitting a PR soon! 🎉

mrharpo:
# Actions: Detail page
Adds custom action buttons to detail page.
- Adds `BaseView` properties:
  -  `exclude_actions_from_detail = []` exclude an action from that view's detail page
  -  `exclude_actions_from_list = []` exclude an action from that view's list page

## Examples
Currently working from our app's feature branch: https://github.com/WGBH-MLA/chowda/pull/158

#### List page
Screen Shot 2023-10-04 at 14 22 28
Optional icons

Screen Shot 2023-10-04 at 14 21 29

#### Detail page
Screen Shot 2023-09-20 at 15 48 23
☝️ Here, "Combine" is excluded from this detail view.

## `actions.py`
Added:
- `action_btn_class`: action button class (default: `btn-outline`)
- `icon_class`: icon class for icon next to button (default: `fa-solid fa-plus`)

## Templates
- Refactors out `action.html` from list page
  - Reused in both list and detail for consistent style 

## JS
- Adds `actions.js` with custom js to apply to `#modal-action`
  - Copied from `list.js`, with minor changes 
- Refactors out `alerts.js` from `list.js`

Closes #301


starlette_admin/actions.py
```diff
@@ -7,8 +7,10 @@ def action(
     name: str,
     text: str,
     confirmation: Optional[str] = None,
-    submit_btn_text: Optional[str] = _("Yes, Proceed"),
+    action_btn_class: Optional[str] = "btn-outline",
     submit_btn_class: Optional[str] = "btn-primary",
+    submit_btn_text: Optional[str] = _("Yes, Proceed"),
+    icon_class: Optional[str] = "fa-solid fa-plus",
```
jowilf, line 13: Can the `icon_class` be made optional? This would prevent any changes in the UI when upgrading.

starlette_admin/views.py
```diff
@@ -222,6 +222,7 @@ class BaseModelView(BaseView):
     exclude_fields_from_detail: Sequence[str] = []
     exclude_fields_from_create: Sequence[str] = []
     exclude_fields_from_edit: Sequence[str] = []
+    exclude_actions_from_detail: Sequence[str] = []
```
jowilf, line 225: We should consider adding an option to exclude from the list as well.
Just a quick thought - you could consider including the flags in the action decorator itself, but it's not a big deal if you decide not to.

starlette_admin/templates/action.html
jowilf, line 1: I think it's better to use a simple dropdown-item for the list page and buttons for the detail page.

starlette_admin/views.py
mrharpo: Yes, I though about adding that, but the PR was already pretty big, and I wanted to make sure the direction was good before starting that. I can add it now to make it complete.

jowilf (commented):
Hello @mrharpo,

Thank you for making the changes; everything looks great!

During testing on my local machine, I noticed one issue. The detail page does not reload after executing an action. As a result, there are inconsistencies between the data in the database and the data displayed in the view. I have some ideas on how to fix this and will do my best to assist you.

starlette_admin/statics/js/actions.js
```diff
@@ -0,0 +1,62 @@
...
+    $("#modal-loading").modal("show");
+    fetch(url, {
+      method: form ? "POST" : "GET",
+      body: form ? new FormData(form) : null,
+    })
+      .then(async (response) => {
+        await new Promise((r) => setTimeout(r, 500));
+        $("#modal-loading").modal("hide");
+        if (response.ok) {
+          successAlert((await response.json())["msg"]);
```
mrharpo, line 24: `actions.js` is copied from `list.js`, with the table elements removed, and a few other changes. The detail page has no equivalent for `table.ajax.reload()`, but it would go here.

Some ideas:
- Refactor `submitAction` to include hooks where needed for detail / list actions
  - Ideal, but could get complicated
- Create an ajax `reload()` function for the detail page
- Refresh the page here, and deliver the success message after reload
- Respond with a redirect to the same page for actions coming from the detail page

Thoughts, @jowilf?

mrharpo: @jowilf I implemented a solution using `localStorage`, where the success message is stored, and removed once the page reloads.

I even extended the pattern to the detail `delete` button, so it shows the success message after redirecting to the list page.

What do you think of this solution?

starlette_admin/statics/js/actions.js
```diff
@@ -0,0 +1,63 @@
+function submitAction(name, form, customResponse) {
```
jowilf, line 1: I think you miss this for actions with customResponse
```suggestion
  $('a[data-no-confirmation-action="true"]').each(function () {
    $(this).on("click", function (event) {
      submitAction(
        $(this).data("name"),
        null,
        $(this).data("custom-response") === true
      );
    });
  });
  
function submitAction(name, form, customResponse) {
```

starlette_admin/statics/js/actions.js
jowilf: It's perfect.

starlette_admin/views.py
```diff
@@ -325,13 +329,28 @@ async def is_action_allowed(self, request: Request, name: str) -> bool:
...
     async def get_all_actions(self, request: Request) -> List[Optional[dict]]:
-        actions = []
         assert self.actions is not None
+        actions = []
         for action_name in self.actions:
             if await self.is_action_allowed(request, action_name):
                 actions.append(self._actions.get(action_name))
         return actions
 
+    async def get_actions(
```
jowilf, line 339: I believe we can extend the logic of get_all_actions(self, request: Request) and eliminate the need for this function.

jowilf: Can we add also a simple unit test [here](https://github.com/jowilf/starlette-admin/blob/main/tests/test_custom_actions.py) to make sure that future PR will not break the `exclude_actions_from_list` and `exclude_actions_from_details` params. I can also help if you authorize me to push here. Thanks

mrharpo: @jowilf I added you as a collaborator to our repo. Feel free to push to this branch as needed!

starlette_admin/statics/js/actions.js
mrharpo: Yes, Thank you!

starlette_admin/views.py
mrharpo: @jowilf I merged `get_actions` into `get_all_actions` and added tests for the excludes.

Good thing, because they caught a logical error that put forbidden actions on the detail page!

jowilf: > @jowilf I added you as a collaborator to our repo. Feel free to push to this branch as needed!
Thanks, but I was thinking about this ->
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork. It also allows the pre-commit bot the pre-commit bot to address linting issues.

jowilf (approved):
Thank you for your contribution🎉. 

There is a weird error on Python 3.8 in the CI that I need to investigate and resolve before merging this.

mrharpo:
The python3.8 tests all passed, but it failed with `Segmentation fault (core dumped)`. 

It might be a transient runner error. Try re-running the job.

starlette_admin/views.py
mrharpo: @jowilf I spent way too long looking for the button before googling it, but that button is not available for forks that live in a GitHub Organization: https://github.com/orgs/community/discussions/5634

You have write permissions for our fork, so if you need to make any further changes, you should be able to.

Issue #335, opened 2023-10-18
Enhancement: Support for SQLAlchemy Models with `column_property` and `hybrid_property`


hasansezertasan:
**Is your feature request related to a problem? Please describe.**
I am getting `TypeError: expected string or bytes-like object, got 'InstrumentedAttribute'` this error code when I try to use `hybrid_property`.

I am getting `AttributeError: Neither 'Label' object nor 'Comparator' object has an attribute 'comment'` this error code when I try to use `column_property`.

These features are looging like not supported yet, any work on them?

**Describe the solution you'd like**
I actually don't know how this system work but I'd like to help with some help 😕 Other quick solution could be something like Flask Admin `column_formatters` property.

**Describe alternatives you've considered**
Flask Admin does support these feature out of the box.

**Additional context**
Great project, keep the good work up!

hasansezertasan:
`hybrid_proprty` support still awaits.

hasansezertasan:
@jowilf `hybrid_proprty` support still awaits.

hasansezertasan:
One fix for `hybrid_property`s is to strictly write the field:

```python
import datetime
from sqlalchemy.ext.hybrid import hybrid_property

from sqlalchemy import create_engine
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column
from starlette.applications import Starlette
from starlette_admin import DateTimeField, IntegerField, StringField
from starlette_admin.contrib.sqla import Admin, ModelView

engine = create_engine("sqlite:///db.sqlite3", connect_args={"check_same_thread": False})


class Base(DeclarativeBase):
    pass


class Mixin:
    id: Mapped[int] = mapped_column(
        primary_key=True,
        autoincrement=True,
        index=True,
        unique=True,
    )
    date_created: Mapped[datetime.datetime] = mapped_column(
        default=datetime.datetime.utcnow,
        index=True,
    )
    date_updated: Mapped[datetime.datetime] = mapped_column(
        default=datetime.datetime.utcnow,
        onupdate=datetime.datetime.utcnow,
        index=True,
    )


class Person(Base, Mixin):
    __tablename__ = "person"
    first_name: Mapped[str]
    last_name: Mapped[str]

    @hybrid_property
    def full_name(self):
        return f"{self.first_name} {self.last_name}"

class PersonView(ModelView):
    fields = [
        IntegerField(
            name="id",
            label="ID",
            help_text="ID of the record.",
            read_only=True,
        ),
[...]

PR #337, opened 2023-10-18
Add support for SQLAlchemy's `column_property`

hasansezertasan:
Adds support for SQLAlchemy's `column_property`

Partially resolves: [335](https://github.com/jowilf/starlette-admin/issues/335)

Open to reccomendations, I might have done it in a place that doesn't suite.


jowilf (commented):
Can you add a simple test for this in [tests/sqla/test_view_meta.py](https://github.com/jowilf/starlette-admin/blob/main/tests/sqla/test_view_meta.py) file?

starlette_admin/contrib/sqla/converters.py
```diff
@@ -123,6 +124,12 @@ class ModelConverter(BaseSQLAModelConverter):
     def _field_common(
         cls, *, name: str, column: Column, **kwargs: Any
     ) -> Dict[str, Any]:
+        if type(column) is Label:
+            return {
+                "name": name,
+                "help_text": name,
```
jowilf, line 130: The `help_text` is only required for columns with a `comment` attribute, so we can add a condition to check for it.

hasansezertasan:
> Can you add a simple test for this in [tests/sqla/test_view_meta.py](https://github.com/jowilf/starlette-admin/blob/main/tests/sqla/test_view_meta.py) file?

I didn't write a test yet and turned it into a draft because I don't think this is the right way to do it. I was trying to detect a `hybrid_propert` column and learnt some stuff. I hope I can finish this utilities as soon as possible.

PR #348, opened 2023-10-21
Feature: Row actions

jowilf:
- [x] Implementation
- [x] Tests
- [x] Documentation

```python
class ArticleView(ModelView):
    ...
    row_actions = ["view", "edit", "go_to_example", "make_published", "delete"]
    row_actions_display_type = RowActionsDisplayType.ICON_LIST # RowActionsDisplayType.DROPDOWN

   @row_action(
        name="make_published",
        text="Mark as published",
        confirmation="Are you sure you want to mark this article as published ?",
        icon_class="fas fa-check-circle",
        submit_btn_text="Yes, proceed",
        submit_btn_class="btn-success",
        action_btn_class="btn-info",
    )
    async def make_published_row_action(self, request: Request, pk: Any) -> str:
        session: Session = request.state.session
        article = await self.find_by_pk(request, pk)
        if article.status == Status.Published:
            raise ActionFailed("The article is already marked as published.")
        article.status = Status.Published
        session.add(article)
        session.commit()
        return "The article was successfully marked as published"

    @link_row_action(
        name="go_to_example",
        text="Go to example.com",
        icon_class="fas fa-arrow-up-right-from-square",
    )
    def go_to_example_row_action(self, request: Request, pk: Any) -> str:
        return f"https://example.com/?pk={pk}"
```
 
#### row_actions_display_type = RowActionsDisplayType.ICON_LIST
 
Screenshot 2023-10-21 at 4 39 58 PM
 
#### row_actions_display_type = RowActionsDisplayType.DROPDOWN

Screenshot 2023-10-21 at 4 48 06 PM

#### detail view

Screenshot 2023-10-21 at 4 52 57 PM

---
Note: Last feature for 0.12.0


jowilf:
Hello @mrharpo, I would love to hear your thoughts on this because it directly affects your work on #302, but i can not see another way to support row actions.

what changed (compared to #302 ):
- Batch actions will only be available in the list
- Row actions are now available on each row and will also appear on the detail page by default

Let me know if you see any issues with this approach! I'm open to suggestions on how to best support the row actions

hasansezertasan:
It looks great, I took a look at the changes, I can't see any problem at the changes you made.

mrharpo:
@jowilf This looks great! 🚀 The implementation doesn't affect our actions directly, but it might be nice if we could merge the concepts, so a task could be available as a `@row_action` and (batch) `@action` without needing to duplicate logic.

In our case, we need to `start` (effectively `publish`) one or many rows from the list page, and also start a row from the detail page. We used the batch `@action` on the detail page as a workaround for a `row_action`, where it sends a list with a single `pk` to the batch action API. However, it would be cleaner if (for example) we used the new `@row_action` instead and allow it to be run as a batch action for multiple records. E.g:

```py
class ArticleView(ModelView):
    ...
    row_actions = ["view", "edit", "go_to_example", "make_published", "delete"]
    batch_actions = ["make_published", "delete"]

   @row_action(
        name="make_published",
        text="Mark as published",
        confirmation="Are you sure you want to mark this article as published ?",
        confirmation_multiple="Are you sure you want to mark these articles as published ?",
        icon_class="fas fa-check-circle",
        submit_btn_text="Yes, proceed",
        submit_btn_class="btn-success",
        action_btn_class="btn-info",
    )
    async def make_published_row_action(self, request: Request, pk: Any) -> str:
    ...
```

By including it in `batch_actions`, it would get run multiple times, returning possibly multiple `ActionFailed` errors in the process, unless some kind of `stop_on_first_failure` flag is set, perhaps.

We're open to suggestions, especially if there's an easier / better way I'm not seeing.

But it's also easy enough for us to include both if it ends up being too complicated, or too much of a breaking change. Again, this is all great, and will work fine for our application.

Related (but definitely a separate issue), we've discussed having some kind of `@view_action` for actions in a view that are unrelated to any records (e.g. `sync_database`).

Thanks for the heads up!

hasansezertasan:
O agree with @mrharpo combination of those logic would be good, maybe gets harder to implement but good. Also, I think we shouldn't remove batch actions from the details view, it's a great feature.

jowilf:
It is possible to use the same method for row actions and batch actions with this simple logic: 
```python
@action(...)
@row_action(...)
async def make_published_action(self, *args) -> str:
  request: Request = args[0]
  if isinstance(args[1], list):
      pks = args[1]
      return "BATCH ACTION"
  else:
      pk = args[1]
      return "ROW ACTION"
```
---
> And also, do we need a link row action? I believe we can annotate the return type and decide the action by inspecting the return type, maybe a redirect response would be enough?

`link_row_action` eliminates the necessity to call the action API. The link is included directly in the `href` attribute of the generated `<a>` element.

hasansezertasan:
> It is possible to use the same method for row actions and batch actions with this simple logic:

That's great to see.

> `link_row_action` eliminates the necessity to call the action API. The link is included directly in the `href` attribute of the generated `<a>` element.

You have thought of everything. Nothing else, great job!

Issue #349, opened 2023-10-21
Enhancement:  configurable logo & skin tenant-based instead of fixed at instantiation time


sglebs:
**Is your feature request related to a problem? Please describe.**

My app is multi tenant. When the user logs in (via vhost detection + OAuth implementation like your auth0 example), I would like to set the app logo at the top/left as something I would read from my the database. Basically make that part multi tenant as well. Currently the logo is defined at the Admin instantiation time:

```
                  logo_url="https://images.com/my.png",
                  login_logo_url="https://images.com/mylogin_logo.png",
```

The login_logo does not impact me because the login goes straight to OAuth.

**Describe the solution you'd like**

The current approach is ok for users not logged in. But once they log in (in my case, with OAuth) then I should be able to set the logo from a column in a database table (loggerUser->company, column for the logo URL).

Please note that the avatar image can be set from the OAUTH profile (you do so in your auth0 example). Maybe the same can be done from there? I mean, when the user logs in, I can call an API to set the image this user will see.

**Describe alternatives you've considered**

Maybe override the template for the top/leftpart (I would have to dig to see where it is) and copy/paste into my project and change it to fetch the logo from the database if the user is logged in, otherwise keep the current behavior.  

Optimization: I could store the URL in the session, so that there is no need to go to the database every time, pretty much like with the avatar image.

**Additional context**

Be able to change the color there would also be quite neat.

Have the documentation explain how to tweak the skin/logo would be quite neat.

hasansezertasan:
Can you try this in your `AuthProvider` subclass?

```python
    def setup_admin(self, admin: BaseAdmin) -> None:
        admin.templates.env.globals["logo_url"] = "logo_url"  # Set logo url dynamically
        return super().setup_admin(admin)
```

You probably need to access the current_user here.

hasansezertasan:
Reading issues really helps!

sglebs:
> Can you try this in your `AuthProvider` subclass?
> 
> ```python
>     def setup_admin(self, admin: BaseAdmin) -> None:
>         admin.templates.env.globals["logo_url"] = "logo_url"  # Set logo url dynamically
>         return super().setup_admin(admin)
> ```
> 
> You probably need to access the current_user here.

Are you suggesting that a user performing login from tenant X should cause a side-effect and change the global admin to use that logo? So whoever logs last wins?  I need the logo to be non-global, not to leak across sessions.

hasansezertasan:
[...]

PR #374, opened 2023-10-31
Adds `AdminConfig` to override app_title, logo_url in the templates.

hasansezertasan:
Check out the `changed files`, there is a good example.

Resolves: #349


jowilf (commented):
I suggest creating another class, `AdminConfig`, and a method `get_admin_config`. What do you think?

hasansezertasan:
> I suggest creating another class, `AdminConfig`, and a method `get_admin_config`. What do you think?

I'm working on it

jowilf (changes requested):
Your changes look great. Please add `get_admin_config` to [auth_provider](https://github.com/jowilf/starlette-admin/blob/main/tests/auth_provider.py) and include a test case for it in [test_auth](https://github.com/jowilf/starlette-admin/blob/main/tests/test_auth.py). Thanks!

jowilf (approved):
Great, Thank you @hasansezertasan

PR #379, opened 2023-11-04
Add SimpleMDEField

hasansezertasan:


starlette_admin/templates/forms/simplemde.html
```diff
@@ -0,0 +1,10 @@
+<div class="{% if error %}{{ field.error_class }}{% endif %}">
+    <textarea name="{{ field.id }}" data-bs-toggle="autosize" class="{{ field.class_ }}
+            {% if error %}{{ field.error_class }}{% endif %}" {{ field.input_params() | safe }}
+            config="{{ field.config() }}"
```
jowilf, line 4: I think we can merge `config` with `field.input_params()` by overriding `input_params()` and adding the `config` attribute. A new form template is not neccessary

jowilf (commented):

starlette_admin/templates/forms/simplemde.html
hasansezertasan: This field is a subclass of TextAreaField and inherits these parameters (attributes to be passed to the textarea tag): `rows`, `minlength`, `maxlength`, `placeholder`, `required`, `disabled`, `readonly`. These values are passed through `input_params` method.

Only `placeholder` attribute still works here. Since the `input_params` method is used to pass `tag attributes` in many places, I believe we shouldn't use it for another purpose.

People might want to use it as it is, as expected, without searching about any configuration about the SimpleMDE in this case.

One last thing, I believe using `input_params` for this will be a little bit "confusing" in the future. Someone may try to pass tag attribute using this but you're recommendation makes it different. I can rename it into `options` if you want.

I think we shouldn't limit the users here but if you insist I can change it.

jowilf: The purpose of the `input_params` is to make it easier to add extra  HTML  attributes to the input, without having to overwrite the form template. This is why many fields share the `input.html` template. 
Here is my suggestion:
```python
class SimpleMDEField(TextAreaField):
    def input_params(self) -> str:
        return super().input_params() + " " + html_params({"data-config": json.dumps(...)})
```

hasansezertasan: > The purpose of the `input_params` is to make it easier to add extra HTML attributes to the input, without having to overwrite the form template.

I'm sorry for my bad English, I was trying to say the same thing lol.




> This is why many fields share the `input.html` template. Here is my suggestion:
> 
> ```python
> class SimpleMDEField(TextAreaField):
>     def input_params(self) -> str:
>         return super().input_params() + " " + html_params({"data-config": json.dumps(...)})
> ```

I just commit some code, can you check them? I did it before seeing your recommendation.

hasansezertasan: 👋 @jowilf I agree with you now. I have made the changes that you requested.

starlette_admin/templates/forms/simplemde.html
jowilf, line 1: I believe we can use or extend 'textarea.html' for this. It would make maintenance easier.

starlette_admin/fields.py
```diff
@@ -276,6 +276,68 @@ def additional_js_links(self, request: Request, action: RequestAction) -> List[s
...
+        self, request: Request, action: RequestAction
+    ) -> List[str]:
+        if action.is_form():
+            return [
+                f"https://cdn.jsdelivr.net/npm/simplemde@{self.version}/dist/simplemde.min.css",
+            ]
+        return []
+
+    def input_params(self) -> str:
+        return html_params({"options": json.dumps(self.options)})
```
jowilf, line 338: It's preferable to use [data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) here

```suggestion
        return html_params({"data-options": json.dumps(self.options)})
```

starlette_admin/fields.py
```diff
@@ -276,6 +276,68 @@ def additional_js_links(self, request: Request, action: RequestAction) -> List[s
...
+    class_: str = "field-simplemde form-control"
+    display_template: str = "displays/simplemde.html"
+    form_template: str = "forms/simplemde.html"
+    version: str = "1.11.2"
+    placeholder: str = ""
+    spell_checker: bool = False
+    status: bool = False
+    hide_icons: List[str] = dc_field(default_factory=list)
+    autofocus: bool = True
+    options: Dict[str, Any] = dc_field(default_factory=dict)
```
jowilf, line 305: ?
```suggestion
    other_options: Dict[str, Any] = dc_field(default_factory=dict)
```

starlette_admin/fields.py
```diff
@@ -276,6 +276,68 @@ def additional_js_links(self, request: Request, action: RequestAction) -> List[s
...
+            self.options["placeholder"] = self.placeholder
+        if self.options.get("spellChecker", None) is None:
+            self.options["spellChecker"] = self.spell_checker
+        if self.options.get("status", None) is None:
+            self.options["status"] = self.status
+        if self.options.get("hideIcons", None) is None:
+            self.options["hideIcons"] = self.hide_icons
+        if self.options.get("autofocus", None) is None:
+            self.options["autofocus"] = self.autofocus
+        super().__post_init__()
```
jowilf, line 319: When using this approach, any updated field attributes after initialization will be ignored. My suggestion is to construct the JSON object in the `html_params` function.

examples/fields/README.md
jowilf, line 1: what about making another PR for the example?

starlette_admin/fields.py
```diff
@@ -276,6 +276,68 @@ def additional_js_links(self, request: Request, action: RequestAction) -> List[s
         return []
 
 
+@dataclass
+class SimpleMDEField(BaseField):
```
jowilf, line 280:
```suggestion
class SimpleMDEField(TextAreaField):
```

starlette_admin/fields.py
hasansezertasan: What about `data_options`, because we are using `data-options` tag attribute?

hasansezertasan:
@jowilf I updated it according to the changes you made in #380.

hasansezertasan:
@jowilf I have renamed the branch so GitHub automatically closed the PR. If it sounds good, I can open another PR.

jowilf:
what about adding this to the library discussed here -> https://github.com/jowilf/starlette-admin/issues/439#issuecomment-1868350268

hasansezertasan:
That could work but I think we should add `CKEditor5Field` to this package because it's one of the most commonly used and most required WYSIWYG editors.

hasansezertasan:
It's available in the [starlette-admin-fields](https://github.com/hasansezertasan/starlette-admin-fields) project.

PR #380, opened 2023-11-04
Refactor TinyMCEEditorField: support custom TinyMCE configuration

hasansezertasan:
I'm not sure if my approach with `config` attribute is the right way to do it but its better than strict and its something...

TinyMCE is quite configurable: Here are some examples: [TinyMCE 6 examples | TinyMCE Documentation](https://www.tiny.cloud/docs/tinymce/6/examples/)


starlette_admin/templates/forms/tinymce.html
```diff
@@ -0,0 +1,10 @@
+<div class="{% if error %}{{ field.error_class }}{% endif %}">
```
jowilf, line 1: I think we can merge `config` with `field.input_params()` by overriding `input_params()` and adding the `config` attribute. A new form template is not neccessary

jowilf (changes requested):

starlette_admin/templates/forms/tinymce.html
hasansezertasan: I think we should keep this conversation here: #379

hasansezertasan:
@jowilf this and #379 follows the same pattern for "configuration". I removed unnecessary templates and used ".data()" to get the "data" attributes, instead of plain `.attr()` method.

hasansezertasan:
The changes you made don't seem to break anything. I was being cautious with all "if-else" things but I think that's the developer's job to be careful about it now.

PR #381, opened 2023-11-04
CKEditor5Field

hasansezertasan:


starlette_admin/fields.py
```diff
@@ -276,6 +276,68 @@ def additional_js_links(self, request: Request, action: RequestAction) -> List[s
         return []
 
 
+@dataclass
+class CKEditor4Field(TextAreaField):
```
jowilf, line 280: CKEditor 4 is no longer supported since June 2023. Continuing to use it may cause compatibility, security, and performance issues.

starlette_admin/templates/forms/ckeditor5.html
```diff
@@ -0,0 +1,10 @@
+<div class="{% if error %}{{ field.error_class }}{% endif %}">
+    <textarea name="{{ field.id }}" data-bs-toggle="autosize" class="{{ field.class_ }}
+            {% if error %}{{ field.error_class }}{% endif %}" {{ field.input_params() | safe }}
+            config="{{ field.config() }}"
```
jowilf, line 4: I think we can merge `config` with `field.input_params()` by overriding `input_params()` and adding the `config` attribute. A new form template is not neccessary

jowilf (changes requested):

starlette_admin/fields.py
hasansezertasan: As far as I know there are people using it and offering this one as an "option" lookgs good but you are right to be concerned. Isn't there any way to enlight people about the risk but at the same time offer this field?

jowilf: I think it's best to avoid the pain of supporting a library that has reached its end of life. Those who still want to use it can create a custom field

hasansezertasan: Alright 🚀

hasansezertasan:
@jowilf I have renamed the branch so GitHub automatically closed the PR. If it sounds good, I can open another PR.

hasansezertasan:
It's available in the [starlette-admin-fields](https://github.com/hasansezertasan/starlette-admin-fields) project.

PR #397, opened 2023-11-14
More generic gitignore

hasansezertasan:
Motivation: I see that [github/gitignore: A collection of useful .gitignore templates](https://github.com/github/gitignore) is quite usefull and common, I think using this as a base would be quite usefull.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ site
 __pycache__/
 .pytest_cache/
 .examples/
+*.sqlite3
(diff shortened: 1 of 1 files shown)


jowilf:
Do we really need this? The current .gitignore seems to be sufficient.

hasansezertasan:
> Do we really need this? The current .gitignore seems to be sufficient for this project

I think this one is more generic and specifically done for Python. It covers many files that generated by editor and operating system and much more.

hasansezertasan:
Closing due to inactivity.

PR #442, opened 2023-12-27
Proposal - Explicit Data Validation: `validate_create` and `validate_edit` instead of just `validate` in `contrib.sqla`

hasansezertasan:
It's a simple draft of what I propose.

Use cases:

- I am creating a book record, the title attribute of the book record can't be changed, so I removed the title attribute from the edit form. Using the same `validate` method for both create and edit processes doesn't help me very much.
- I have a model that is quite complex with over 50 attributes. I kept the create form minimal and it has only some basic attributes to get started. The edit form has all 50 attributes.

**The validation process is not the same in both create and edit processes for these scenarios.**

I had these scenarios in my projects and I have overridden the `create` and `edit` methods.

I think if we are going to have a `validate` method at the end of the day, we better have two separate validate methods for both `create` and `edit` functionality.

@jowilf If this sounds good, I can keep working on it.

### DIFF
--- a/starlette_admin/contrib/sqla/view.py
+++ b/starlette_admin/contrib/sqla/view.py
@@ -480,6 +480,7 @@ class ModelView(BaseModelView):
         """
         Inherit this method to validate your data before creating.
         """
+
     async def validate_edit(self, request: Request, data: Dict[str, Any]) -> None:
         """
         Inherit this method to validate your data before editing.
--- a/starlette_admin/contrib/sqla/view.py
+++ b/starlette_admin/contrib/sqla/view.py
@@ -476,10 +476,19 @@ class ModelView(BaseModelView):
 
         """
 
+    async def validate_create(self, request: Request, data: Dict[str, Any]) -> None:
+        """
+        Inherit this method to validate your data before creating.
+        """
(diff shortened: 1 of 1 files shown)


jowilf:
Have you considered checking the `request.state.action` attribute? For example:
```py
    async def validate(self, request: Request, data: Dict[str, Any]) -> None:
        if request.state.action == RequestAction.CREATE:
            ...
        else: # RequestAction.EDIT
            ...
```

hasansezertasan:
> Have you considered checking the `request.state.action`? For example:

That solution didn't come to my mind but we could have done the same thing, instead of two different methods like `before_create` or `before_edit`. Why didn't we?

We could have just one method and use `request.state.action`... Something like this?

```python
    async def before(self, request: Request, data: Dict[str, Any]) -> None:
        if request.state.action == RequestAction.CREATE:
            ...
        else: # RequestAction.EDIT
            ...
```

I know it's not the same but I think if we are meant to have two separate `before_` methods for both `create` and `edit`, we might also have two separate `validate_` methods for both `create` and `edit`.

In the end, what's the downside? In my point of view, it only gives a better developer experience.

hasansezertasan:
Closing due to inactivity.

Issue #433, opened 2023-12-23
Enhancement: APScheduler Integration


hasansezertasan:
**Is your feature request related to a problem? Please describe.**
APScheduler is a "Task scheduling library for Python". It allows us to schedule some tasks, mostly based on functions.
The problem is this: ability to view, add, reschedule, cancel or stop a task/job from the Starlette Admin interface.

**Describe the solution you'd like**
I'm working on something.

**Describe alternatives you've considered**
I didn't consider but FastAPI Amis Admin has support for it as third party solution: [amisadmin/fastapi-scheduler: FastAPI-Scheduler is a simple scheduled task management FastAPI extension based on APScheduler.](https://github.com/amisadmin/fastapi-scheduler)

**Additional context**
I believe this could be a seperate package like `fastapi-scheduler` but still, I think it should be in this package with optional dependencies.

hasansezertasan:
I will address the same issue for Rocketry in the future.

Issue #389, opened 2023-11-08
Enhancement: `expose` decorator for Flask Admin


hasansezertasan:
Flask Admin had this one for a long time and it is quite usefull I think Starlette Admin should have them too.

But it lacks of things like adding the exposed routes to the menu.

jowilf:
a new decorator @route was added in #798

PR #465, opened 2024-01-08
Feature: `expose` decorator from Flask Admin

hasansezertasan:
Related to #389, still in progress, open to reviews and recommendations.

TODO:

- [ ] Complete the `expose` logic.
- [ ] Turn every existing route into an exposed route.

Gains:

- Implementing our routes in a more `starlette-admin` environment.
- IMHO, More flexible and customizable: We won't have to override "BaseAdmin" again and again for different views that require different logic.
- IMHO, Makes it easier to implement something like #433 and `FormView`.
- Makes it flexible to use it with other frameworks like FastAPI. Meaning: a dedicated route for every endpoint with dedicated responsibilities. `"/api/{identity}"`, won't take "identity" as a path parameter it will be fixed for each view. This will make Swagger implementation easier.
- Makes it easier to implement features mentioned at #411, #412, and #413

Example:

```python
from typing import Any

from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import PlainTextResponse
from starlette_admin import CustomView, BaseAdmin
from starlette_admin.expose import expose


class IndexView(CustomView):
    @expose("/index", methods=["GET"])
    async def index(self, request: Request) -> Any:
        return PlainTextResponse("Hello, world!")


app = Starlette()
admin = BaseAdmin(title="@expose example")
admin.add_view(IndexView("Index"))
admin.mount_to(app)

```


hasansezertasan:
Closing due to inactivity.

PR #469, opened 2024-01-13
Refactor `form.js` to allow custom field initializer callbacks

jowilf:


hasansezertasan:
@jowilf I want to ask you something, do we have a way to watch the entire documentation and initialize some JS when an element with a certain selector is created? For example, trigger `registerFieldInitializer` when an element is created with a class containing "field".

jowilf:
I'm not sure if I fully understand your question but if you create an element you need to call `initializeFields` on the element. This function will loop through all the registered callbacks and invoke them. Take a look at the ListField implementation that adds dynamically a field template  ->  https://github.com/jowilf/starlette-admin/blob/63948e34919d62556c66266e67a3f3ecdfa7af8d/starlette_admin/statics/js/form.js#L210
You don't have to watch the entire document if you can control when the element is created.

hasansezertasan:
> I'm not sure if I fully understand your question but if you create an element you need to call `initializeFields` on the element.

Let me try to explain one more time. Do you know anything about HTMX? I'm using it in my project.

Here is an example of what I am doing...

[</> htmx ~ Examples ~ Modal Dialogs with UIKit](https://htmx.org/examples/modal-uikit/). Navigate to this page and click "Show Modal Dialog". The modal and its content (a form) are fetched from the backend. So the backend returns an HTML, and HTMX puts it where we tell it to do.

Let's say that one of the inputs in that form is a JavaScript-initialized field, maybe it is supposed to be a TinyMCE Editor and it has this class: "field-tinymce-editor form-control".

With HTMX, I can not trigger a function based on the content of the response and I don't want to put an extra "script" tag into that response HTML, the JavaScript file is already fetched, so why fetch it again?

hasansezertasan:
> I'm not sure if I fully understand your question but if you create an element you need to call `initializeFields` on the element.

Let me try to explain one more time. Do you know anything about HTMX? I'm using it in my project.

Here is an example of what I am doing...

[</> htmx ~ Examples ~ Modal Dialogs with UIKit](https://htmx.org/examples/modal-uikit/). Navigate to this page and click "Show Modal Dialog". The modal content (a form) is fetched from the backend. So the backend returns an HTML, and HTMX puts it where we tell it to put.

Let's say that one of the inputs in that form is a JavaScript-initialized field, maybe it is supposed to be a TinyMCE Editor and it has this class: "field-tinymce-editor form-control". With HTMX, I can not trigger a function based on the content of the response and I don't want to put an extra "script" file loader into that response HTML, the JavaScript file is already fetched, so why fetch it again?

So what I am asking is this, is there any way to "watch" the document and trigger a function (Field initializer) when an element is created with a certain selector, in this example: `.field-tinymce-editor`.

Even if I can not trigger a function based on the content of the response and I don't want to put an extra "script" file loader into that response, I can still add `<script>initializeFields(document);</script>` (or something like that). But I want it to be as dynamic as possible and I believe there is only one way to make it: watch the document for the creation of an element with a selector.

I hope this one was clear enough.

hasansezertasan:
I have found these two JS packages:

[kubetail-org/sentineljs: Detect new DOM nodes using CSS selectors (650 bytes)](https://github.com/kubetail-org/sentineljs)

[uzairfarooq/arrive: Watch for DOM elements creation and removal](https://github.com/uzairfarooq/arrive)

They are quite helpful for this purpose: watch the DOM and do something when a target element is created.

PR #472, opened 2024-01-15
Add support for record based `row-action` authorization

hasansezertasan:
Related to #463, still in progress, open to reviews and recommendations.

The logic is simple, pass `obj` to the `is_row_action_allowed` method but I'm struggling.

Issue #463, opened 2024-01-05
Question: How do we modify available actions for each item?


hasansezertasan:
How do we modify available actions for each item?

My question is based on this StackOverflow question: [How to modify available actions for each item in a list in Flask-Admin - Stack Overflow](https://stackoverflow.com/questions/63291348/)

Consider this SQLAlchemy model:

```python
class Record(Base):
    id: Mapped[int]
    attr: Mapped[str]
    can_delete: Mapped[bool]
    can_edit: Mapped[bool]
    can_view_details: Mapped[bool]
    can_do_this: Mapped[bool]  # `@row_action`
    can_do_that: Mapped[bool]  # `@link_row_action`
```

The situation here is that not every record can be deleted, edited, or viewed. Some items can be deleted yes but some items can not, so what to do here? Disable the "delete" row and batch actions for the entire list or disable it for only relevant records.

jowilf:
You can override the [row-actions.html](https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/templates/row-actions.html) template for complex use case

hasansezertasan:
> You can override the [row-actions.html](https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/templates/row-actions.html) template for complex use cases

Can you please give an example with [this SQLAlchemy model](https://github.com/jowilf/starlette-admin/issues/463#issue-2066745161)

hasansezertasan:
> You can override the [row-actions.html](https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/templates/row-actions.html) template for complex use cases

I tried it but there is no `obj` in the `row-actions.html` template. Can you please give an example?

jowilf:
> I tried it but there is no obj in the row-actions.html template. Can you please give an example?

You have access to the `request` and the primary key (`pk`).

hasansezertasan:
> > I tried it but there is no obj in the row-actions.html template. Can you please give an example?
> 
> You have access to the `request` and the primary key (`pk`).

Do you have an example in mind? It's a three weeks old issue and I wanted to address it since it'll most probably be a feature on demand.

jowilf:
I was giving you some hints to implement the solution on StackOverflow.

You can create a custom method in your ModelView:

```py
class MyModelView(ModelView):

    def allow_row_action(self, action, pk)-> bool:
        return True
```

And in your custom row_actions.html:

```j2
{% for action in _actions -%}
[...]

Issue #444, opened 2023-12-27
Question: How to validate a @row_action or @action with `form` attribute?


hasansezertasan:
As I said in #440, I'm using `row_actions` to create child-models. I construct the `@row_action` with `form` parameter and some of the inputs are required and some of them have attributes like `min=2023`. When I click the `Yes, Proceed` button, the form is submitted without satisfying the form requirements, let's say leaving an input empty even if it is required, the modal just disappears without validating the form.

**Environment (please complete the following information):**
 - Starlette-Admin version: [e.g. 0.12.2]
 - ORM/ODMs: [SQLAlchemy]

jowilf:
I think we need to [check the validity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/checkValidity) of the form before submitting.

hasansezertasan:
I'll check it if I can come up with a solution, keep in mind, that I'm not strong at JavaScript.

hasansezertasan:
I added my modifications below:

https://github.com/jowilf/starlette-admin/blob/769a85ffdcc3fbf0b2e9dff235678e88fa4128b4/starlette_admin/statics/js/actions.js#L42-L68

as

```javascript
  initActionModal() {
    let self = this;
    $("#modal-action").on("show.bs.modal", function (event) {
      let button = $(event.relatedTarget); // Button that triggered the modal
      let confirmation = button.data("confirmation");
      let form = button.data("form");
      let name = button.data("name");
      let submit_btn_text = button.data("submit-btn-text");
      let submit_btn_class = button.data("submit-btn-class");
      let customResponse = button.data("custom-response") === true;
      let isRowAction = button.data("is-row-action") === true;

      let modal = $(this);
      modal.find("#actionConfirmation").text(confirmation);
      let modalForm = modal.find("#modal-form");
      modalForm.html(form);
      let actionSubmit = modal.find("#actionSubmit");
      actionSubmit.text(submit_btn_text);
      actionSubmit.removeClass().addClass(`btn ${submit_btn_class}`);
      actionSubmit.unbind();
      actionSubmit.on("click", function (event) {
        const formElements = modalForm.find("form");
        const form = formElements.length ? formElements.get(0) : null;
        if (form && !form.checkValidity()) {
          event.preventDefault();
          event.stopPropagation();
          console.log("Form is not valid but modal is closed anyway!")
          return;
        }
        else if (form) {
          console.log("Form is valid!")
[...]

PR #505, opened 2024-02-13
Ability to use Jinja2 templates in action forms

geoolekom:
This PR allows the use of Jinja2 templates in action forms, making it more flexible than plain HTML.

- Moved form initialization to HTML from JS to use Jinja2 for rendering.
- Separated modal windows for each action. This is also allows to use `select2` controls the same way it is done in editing/creation forms, without duplicating logic and without re-creating controls.
- Added a base form template that will be rendered if you pass form_fields to the rendering context.
- Separated the action and request_action logic so that these don't get mixed up in different parts of the code.
- Added an example of use and documented the feature.

### DIFF
--- a/starlette_admin/actions.py
+++ b/starlette_admin/actions.py
@@ -1,6 +1,8 @@
-from typing import Any, Awaitable, Callable, Optional
+from typing import Any, Awaitable, Callable, Optional, Sequence
 
 from starlette_admin.i18n import lazy_gettext as _
+
+from .fields import BaseField
 
 
 def action(
@@ -11,6 +13,8 @@ def action(
     submit_btn_text: Optional[str] = _("Yes, Proceed"),
     icon_class: Optional[str] = None,
     form: Optional[str] = None,
+    form_template: Optional[str] = "forms/action_form.html",
+    form_fields: Optional[Sequence[BaseField]] = None,
     custom_response: Optional[bool] = False,
 ) -> Callable[[Callable[..., Awaitable[str]]], Any]:
     """
@@ -26,6 +30,8 @@ def action(
                 `btn-outline-danger`, ...)
         icon_class: Icon class (ex. `fa-lite fa-folder`, `fa-duotone fa-circle-right`, ...)
         form: Custom form to collect data from user
+        form_template: Custom action form template file
+        form_fields: Fields of the form template
         custom_response: Set to True when you want to return a custom Starlette response
             from your action instead of a string.
 
@@ -91,6 +97,8 @@ def action(
             "submit_btn_class": submit_btn_class,
             "icon_class": icon_class,
             "form": form if form is not None else "",
+            "form_template": form_template,
+            "form_fields": form_fields or [],
             "custom_response": custom_response,
         }
         return f
@@ -107,6 +115,8 @@ def row_action(
     submit_btn_text: Optional[str] = _("Yes, Proceed"),
     icon_class: Optional[str] = None,
     form: Optional[str] = None,
+    form_template: Optional[str] = "forms/action_form.html",
+    form_fields: Optional[Sequence[BaseField]] = None,
     custom_response: Optional[bool] = False,
     exclude_from_list: bool = False,
     exclude_from_detail: bool = False,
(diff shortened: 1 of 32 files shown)


geoolekom:
This also partially solves the problem described in https://github.com/jowilf/starlette-admin/pull/375.
This allows fields to be used for rendering, but not for validation.

starlette_admin/actions.py
```diff
@@ -11,6 +11,8 @@ def action(
     submit_btn_text: Optional[str] = _("Yes, Proceed"),
     icon_class: Optional[str] = None,
     form: Optional[str] = None,
+    form_template: Optional[str] = "forms/action_form.html",
+    form_context: Optional[dict] = None,
```
hasansezertasan, line 15:
```suggestion
    form_context: Optional[Dict] = None,
```

To support older Python versions, you should import Dict from typing or typing_extensions and use `Dict` type.

geoolekom:
> Edit: At #444, we talked about that validation issue. I recommend checking it. It would be awesome if you could integrate validation feature to the `ActionManager` API 🙏.

I believe that client validation is not enough, and furthermore, it cannot be trusted. I can suggest a temporary workaround to validate forms, which you may see in examples: 

https://github.com/jowilf/starlette-admin/pull/505/files#diff-020e73342e8e6a1a797cfe0dbdaaeb533146e3e57e8373a9059a6e5b4855f55eR85

```python
    async def add_tag(self, request: Request, pks: List[Any]):
        form_fields = [HasOne("tag", identity="tag")]
        try:
            form = (
                await request.form()
            )  # from starlette_admin.base.BaseAdmin._render_edit
            data = {
                f.name: f.parse_form_data(request, form, request.state.action)
                for f in form_fields
            }  # from starlette_admin.base.BaseAdmin.form_to_dict
        except Exception as e:
            raise ActionFailed(f"Form processing error: {e}") from e

        data = await self.arrange_data_for_fields(request, data, form_fields)
        await self.validate_add_tag(request, data)
```

Of course, I would like to be able to hide all this boilerplate code inside `handle_action` and `handle_row_action`, and have the validated form as one of the action arguments. But this would require significant refactoring.

I would like to be able to validate the form in the same way as in, for example, `starlette_admin.contrib.sqla.view.ModelView.create`, but the logic of this validation is partly contained in BaseAdmin, partly in View, partly in ModelView. Therefore, I would leave this refactoring for later, and now limit myself to only processing `FormValidationError`.

geoolekom:
> But this would require significant refactoring.

Taking a fresh look, I realized that there wasn't much refactoring. Added form parsing in BaseView. Now, if the action has `form_fields` specified, then the function must take three arguments, the third of which is data.

Also, I realized that module-level `form_context` makes no sense. I can't think of a single example in which something is already defined at the module-level, but it cannot be specified directly in a custom `form_template`.

I left fixups to simplify your review. Will squash them later after you will check the changes.

hasansezertasan:
> I would like to be able to validate the form in the same way as in, for example, `starlette_admin.contrib.sqla.view.ModelView.create`, but the logic of this validation is partly contained in BaseAdmin, partly in View, partly in ModelView. Therefore, I would leave this refactoring for later, and now limit myself to only processing `FormValidationError`.

Well, that separation bothers me a lot...

> So, this solves the problem stated at #444

But the login is in the backend instead of frontend right?

> Also, I realized that module-level `form_context` makes no sense. I can't think of a single example in which something is already defined at the module-level, but it cannot be specified directly in a custom `form_template`.

I totally agree with you, if form_context is going to take one keyword (form_fields), why not just use that keyword?

> I left fixups to simplify your review. Will squash them later after you will check the changes.

I'll take a look at it as soon as possible, great work!

geoolekom:
> But the login is in the backend instead of frontend right?

I don't see anything about login at #444. It's a typo? However, validation is completely in the backend, and it is completely similar to how it is done in `create`, except that the `validate_*` function must be called explicitly.

> I'll take a look at it as soon as possible, great work!

Thank you, looking forward to your review!

hasansezertasan:
> I don't see anything about login at #444. It's a typo?

Yep, it's a typo. I was walking about validation logic.

hasansezertasan (approved):
Well, you've done a great job 👏.

I cloned your fork and tried the example. It seems to me that everything is working as it supposed to be but I don't think my approval will (and should not) be enough to be merged. Let's hear what @jowilf has to say about this, as for me, it's far better implementation than #375.

I'll definetely use the changes you made. 🚀.

geoolekom:
> What do you think about https://github.com/jowilf/starlette-admin/issues/463 and https://github.com/jowilf/starlette-admin/pull/472?

I will look into these issues and write my comment as soon as possible. Since you have already did the review, I will squash the fixups.

geoolekom:
@jowilf @hasansezertasan Any updates on this?

jowilf:
support added in https://jowilf.github.io/starlette-admin/user-guide/actions/#dynamic-action-forms

PR #506, opened 2024-02-13
DOCS: correct the default templates names.

omarmoo5:


jowilf (approved):
Thanks @omarmoo5, good catch!

PR #508, opened 2024-02-14
Get primary key fields from mapper

### LINKED ISSUE #507: Bug:  dictionary changed size during iteration
Somehow my models aren't completely loaded when the admin is initialized. During the setup of the admin fields are added to the model and therefore `self.model.__dict__` changed size during iteration.

Fixing the problem is easier than isolating a test case, so I do not have something to reproduce the error. 

 - Starlette-Admin version: 0.13.2
 - ORM/ODMs: SQLAlchemy

foarsitter:
Caused a lot of pain to fix #507 

-  `Mapper.primary_key` has no relation with the actual attribute of the class 
-  `Mapper._primary_key_propkeys` isn't ordered so I had to do that by myself


foarsitter:
#507 is fixed so this PR is no longer relevant. Thanks @jowilf!

Issue #552, opened 2024-06-25
Bug: UUID pk gives JSON serialization error when excluded from the list


alg:
**Describe the bug**
When I have a model with the pk (id) field of UUID type and include it in the list, it works fine. But if I exclude it, while leaving among the fields, it gives an error:

```
  File "/Users/alg/p/template-fastapi/.venv/lib/python3.11/site-packages/starlette/responses.py", line 187, in render
    return json.dumps(
           ^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.9/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type UUID is not JSON serializable
```

**To Reproduce**
- Create a model with the id field of type UUID.
- List it among fields in the model.
- Exclude it from the list.

**Environment (please complete the following information):**
 - Starlette-Admin version: 0.14.0
 - ORM/ODMs: SQLAlchemy

barbarrista:
Hi! It's works for me

```python
class UuidSerializableModelView(ModelView):
    async def get_pk_value(
        self,
        request: Request,
        obj: Any,  # noqa: ANN401
    ) -> Any:  # noqa: ANN401
        pk = await super().get_pk_value(request, obj)
        if isinstance(pk, UUID):
            return str(pk)
        return pk

class MyView(UuidSerializableModelView):
    ...
```

alg:
[...]

PR #553, opened 2024-06-28
Fix JSON serialization error for UUID primary keys when excluded from list

alg:
Addresses #552 

When PK is not listed among fields it needs proper serialization in case it's not of a JSON-serializable type already (for example, UUID).

- `starlette_admin.BaseModelView` returns the PK value unchanged (for backwards compatibility)
- `starlette_admin.contrib.sqla.ModelView` uses PK field for proper serialization

### DIFF
--- a/starlette_admin/views.py
+++ b/starlette_admin/views.py
@@ -719,8 +719,8 @@ class BaseModelView(BaseView):
                     obj_serialized[field.name] = None
                 elif isinstance(field, HasOne):
                     if action == RequestAction.EDIT:
-                        obj_serialized[field.name] = await foreign_model.get_pk_value(
-                            request, value
+                        obj_serialized[field.name] = (
+                            await foreign_model.get_serialized_pk_value(request, value)
                         )
                     else:
                         obj_serialized[field.name] = await foreign_model.serialize(
@@ -729,7 +729,7 @@ class BaseModelView(BaseView):
                 else:
                     if action == RequestAction.EDIT:
                         obj_serialized[field.name] = [
-                            (await foreign_model.get_pk_value(request, obj))
+                            (await foreign_model.get_serialized_pk_value(request, obj))
                             for obj in value
                         ]
                     else:
@@ -750,11 +750,14 @@ class BaseModelView(BaseView):
                 "result": await self.select2_result(obj, request),
             }
         obj_meta["repr"] = await self.repr(obj, request)
+
+        # Make sure the primary key is always available
+        pk_attr = not_none(self.pk_attr)
+        if pk_attr not in obj_serialized:
+            pk_value = await self.get_serialized_pk_value(request, obj)
+            obj_serialized[pk_attr] = pk_value
+
         pk = await self.get_pk_value(request, obj)
-        obj_serialized[not_none(self.pk_attr)] = obj_serialized.get(
-            not_none(self.pk_attr),
-            pk,  # Make sure the primary key is always available
-        )
         route_name = request.app.state.ROUTE_NAME
         obj_meta["detailUrl"] = str(
             request.url_for(route_name + ":detail", identity=self.identity, pk=pk)
@@ -878,6 +881,23 @@ class BaseModelView(BaseView):
 
     async def get_pk_value(self, request: Request, obj: Any) -> Any:
         return getattr(obj, not_none(self.pk_attr))
+
+    async def get_serialized_pk_value(self, request: Request, obj: Any) -> Any:
+        """
+        Return serialized value of the primary key.
+
+        !!! note
+
+            The returned value should be JSON-serializable.
+
+        Parameters:
+            request: The request being processed
+            obj: object to get primary key of
+
+        Returns:
+            Any: Serialized value of a PK.
+        """
+        return await self.get_pk_value(request, obj)
 
     def _length_menu(self) -> Any:
         return [
--- a/tests/sqla/test_view_serialization.py
+++ b/tests/sqla/test_view_serialization.py
@@ -1,0 +1,92 @@
+import uuid
+
+import pytest
+import pytest_asyncio
+from httpx import AsyncClient
+from sqlalchemy import Boolean, Column, ForeignKey, String
+from sqlalchemy.engine import Engine
+from sqlalchemy.orm import Session, declarative_base, relationship
+from starlette.applications import Starlette
+from starlette_admin.contrib.sqla import Admin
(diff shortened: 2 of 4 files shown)


tests/sqla/test_view_serialization.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Ensures PK is present in the serialized data and properly serialized as a string.
+    """
+
+    user_id = uuid.uuid1()
+    user = User(id=user_id, name="Jack")
+
+    data = await user_view.serialize(user, req, RequestAction.LIST)
+
+    assert data["id"] == str(user_id)
```
jowilf, line 71: Starlette recommends [testclient](https://www.starlette.io/testclient/) to avoid creating the request object. I think It's better to simply call the listing API ,which will, in turn, invoke the serialize method. Here is an example -> https://github.com/jowilf/starlette-admin/blob/41a59ccdc61e0c5ce88f0a34436cbd0fb8d3309f/tests/sqla/test_sync_engine.py#L177

starlette_admin/views.py
```diff
@@ -750,11 +750,14 @@ async def serialize(
                 "result": await self.select2_result(obj, request),
             }
         obj_meta["repr"] = await self.repr(obj, request)
+
+        # Make sure the primary key is always available
+        pk_attr = not_none(self.pk_attr)
+        if pk_attr not in obj_serialized:
```
jowilf, line 756: 👍

jowilf (commented):
Thanks for the PR, i added some comments. can you take a look?

starlette_admin/views.py
```diff
@@ -750,11 +750,14 @@ async def serialize(
...
+            pk_value = await self.get_serialized_pk_value(request, obj, action)
```
jowilf, line 757: We need to use `get_serialized_pk_value` at these lines as well.
https://github.com/jowilf/starlette-admin/blob/8c6f58e27c76649a78feb92abd5efcdd56636d0e/starlette_admin/views.py#L722

https://github.com/jowilf/starlette-admin/blob/8c6f58e27c76649a78feb92abd5efcdd56636d0e/starlette_admin/views.py#L732

tests/sqla/test_view_serialization.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.fixture
+def user_view():
+    return UserView(User)
+
+
+@pytest.fixture
+def app(user_view: UserView):
+    app = Starlette()
+    app.state.ROUTE_NAME = "admin"
```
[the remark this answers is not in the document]
alg: Yes, it's required.

tests/sqla/test_view_serialization.py
alg: This poses the question, why the serialization is dependent on the HTTP request object? Maybe consider decoupling them via repackaging parts of the request the serialization is dependent on into a separate DTO (the serialization context). This would give you exactly what I've done -- testing serialization thoroughly without executing fake HTTP requests.

I'll see how doing it through testclient looks for now.

jowilf: The request is used to generate some metadata as you can see here  https://github.com/jowilf/starlette-admin/blob/13e902fc06e1a68c0b2ba88b765aac569ba2aa4a/starlette_admin/views.py#L760

 > Maybe consider decoupling them via repackaging parts of the request the serialization is dependent on into a separate DTO (the serialization context).

This a good idea, can you open a discussion to present your approach?

alg: Hm, I'm seeing your commits on this. I was cleaning it up when seen them. )

alg:
@jowilf here's the gist of my changes that are now conflicting with yours -- https://gist.github.com/alg/b03907a4edcd44fc5adbfd27c90fcf76

Let me know how you want to deal with it. Are you taking over, or me integrating your changes?

tests/sqla/test_view_serialization.py
jowilf: Hi, sorry I thought you were struggling with the test, and  I was trying to help. your implementation with the platform-independent UUID is better, the only thing missing is the relationship (for `foreign_model.get_ser...`) that you can copy and paste from my commit. Thanks

alg: Oh, ok. I was just busy yesterday. Will finish the PR changes today.

tests/sqla/test_view_serialization.py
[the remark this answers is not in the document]
alg: Removed custom request creation

alg:
@jowilf Pushed the updates to tests incorporating your changes. Please review.

alg:
I guess, I need to provide the UUID for MySQL now for the tests of serialization to pass. :) See what I meant? Will have a look.

alg:
Fixed the issue with tests and verified locally by running tests for MySQL.

tests/sqla/utils.py
```diff
@@ -41,3 +44,32 @@ def get_test_container(name: str) -> Container:
...
+
+    def load_dialect_impl(self, dialect):
+        return mysql.CHAR(32) if dialect == "mysql" else types.CHAR(32)
+
+    def process_bind_param(self, value, dialect):
+        if value is None:
+            return value
+
+        if not isinstance(value, uuid.UUID):
+            value = uuid.UUID(value)
```
jowilf, line 64:
```suggestion
```
I think this condition is not necessary

tests/sqla/utils.py
```diff
@@ -41,3 +44,29 @@ def get_test_container(name: str) -> Container:
...
+    """
+
+    impl = types.CHAR
+
+    def load_dialect_impl(self, dialect):
+        return mysql.CHAR(32) if dialect == "mysql" else types.CHAR(32)
+
+    def process_bind_param(self, value, dialect):
+        if value is None:
+            return value
```
jowilf, line 61:
```suggestion
```
This is not necessary for the current test logic

tests/sqla/utils.py
```diff
@@ -41,3 +44,29 @@ def get_test_container(name: str) -> Container:
...
+
+        return value.hex
+
+    def process_result_value(self, value, dialect):
+        if value is None:
+            return value
```
jowilf, line 67:
```suggestion
```
This is not necessary for the current test logic

jowilf (approved):
Thanks for the PR @alg 🎉

PR #573, opened 2024-08-28
Interval field

regisin:
Handles #572


starlette_admin/fields.py
```diff
@@ -1267,3 +1272,57 @@ def additional_css_links(
...
+    form_template: str = "forms/interval.html"
+    display_template: str = "displays/interval.html"
+
+    async def parse_form_data(
+        self, request: Request, form_data: FormData, action: RequestAction
+    ) -> Any:
+        timedelta_params = {
+            "weeks": (
+                0
+                if form_data.get(f"{self.name}_weeks") == ""
```
jowilf, line 1289: You should use `self.id` here so that the field can be used with ListField and CollectionField.
```suggestion
                if form_data.get(f"{self.id}_weeks") == ""
```

jowilf: same for other attributes

starlette_admin/helpers.py
```diff
@@ -151,3 +152,34 @@ def not_none(value: Optional[T]) -> T:
     if value is not None:
         return value
     raise ValueError("Value can not be None")  # pragma: no cover
+
+
+def timedelta_to_components(td: timedelta):
```
jowilf, line 157: can we some unit tests for this method?

starlette_admin/statics/js/render.js
```diff
@@ -111,4 +111,20 @@ const render = {
       )
       .join("")}</div>`;
   },
+  interval: function render(data, type, full, meta, fieldOptions) {
+    if (!data) return null_column();
+    if (Array.isArray(data) && data.length == 0) return empty_column();
+    data = Array.isArray(data) ? data : [data];
+    if (type != "display") return data.map((d) => new URL(d.url));
```
jowilf, line 118: This should be a string representation of the data 
```suggestion
    if (type != "display") return data.map((d) => );
```

starlette_admin/helpers.py
```diff
@@ -151,3 +152,34 @@ def not_none(value: Optional[T]) -> T:
...
+    total_seconds -= weeks * seconds_in_week
+
+    days = td.days % 7
+    hours = td.seconds // seconds_in_hour
+    minutes = (td.seconds % seconds_in_hour) // seconds_in_minute
+    seconds = td.seconds % seconds_in_minute
+    milliseconds = td.microseconds // 1000
+    microseconds = td.microseconds % 1000
+
+    return {
```
jowilf, line 177: I would prefer using [format_timedelta](https://babel.pocoo.org/en/latest/dates.html#time-delta-formatting) from babel) and extend `StringField` to skip the display and render template.

starlette_admin/templates/forms/interval.html
```diff
@@ -0,0 +1,33 @@
+<div class="{% if error %}{{ field.error_class }}{% endif %}">
```
jowilf, line 1: I would prefer a view like Dimension in this form https://starlette-admin-demo.jowilf.com/admin/mongoengine/product/create. You can copy form https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/templates/forms/collection.html

jowilf (changes requested):
Hi thank you for the PR, l left some few comments.

Also can we add a converter for Interval in sqlalchemy and timedelta for Odmantic

starlette_admin/helpers.py
regisin: There's a loss of precision with that though, for example (delta is 1 week and 5 hours):

```
>>> delta
datetime.timedelta(days=7, seconds=18000)
>>> format_timedelta(delta, locale='en_US', granularity='hour')
'1 week'
```

It all depends on the use case of course, but for me I'm making some cron job stuff, and 5 hours make a lot of difference.. if internally the value is precise, the displayed version should reflect that in my opinion.

What do you think? I could return a `str` from the helper function and skip the template if needed.

regisin: Found this: https://humanize.readthedocs.io/en/stable/time/#humanize.time.precisedelta

Would that work?

regisin: There's something I'm confused about. The `serialize_value` returns a value that the form can pull existing data from, right? Is this method also used for creating a string representation of the data, such as with babel or humanize?

If that's the case, then I don't know how to read the info from `data` in the form, as it is now a string...

starlette_admin/templates/forms/interval.html
regisin: Hey, I'm not sure what to do with these converters..what do they do? I just mimic'd the "Time" one:

```python
    @converts("Interval")
    def conv_interval(self, *args: Any, **kwargs: Any) -> BaseField:
        return IntervalField(
            **self._field_common(*args, **kwargs),
        )
```

Is that it?

jowilf: Yes, that's it. The converter's role is to automatically find the appropriate field for each attribute. For SQLAlchemy, it uses the attribute type. Think of it as a map between the attribute type and the corresponding field.

starlette_admin/helpers.py
jowilf: The benefit of [format_timedelta](https://babel.pocoo.org/en/latest/dates.html#time-delta-formatting) is the support for internationalization. The way humanize handles internationalization will lead to race conditions when handling multiple requests.

One way is to find all the components and call `format_timedelta` on each of them and then join with a separator(space) 

> If that's the case, then I don't know how to read the info from data in the form, as it is now a string...

(you probably already figured it out) you can return different values based on the request action `request.state.action`

starlette_admin/templates/forms/interval.html
regisin: Updated the pr.

jowilf (changes requested):
it looks good just some minor comments

starlette_admin/fields.py
```diff
@@ -18,10 +18,16 @@
     Union,
 )
 
+from babel.dates import format_timedelta
```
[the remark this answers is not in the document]
jowilf: Yes, the user is not able to use it if the import fails, but they are required to install it before using the field. The idea is to make the arrow package optional, so users who don't want to use ArrowField are not forced to install it.

starlette_admin/contrib/odmantic/converters.py
```diff
@@ -109,6 +110,10 @@ def conv_bson_int64(self, *args: Any, **kwargs: Any) -> BaseField:
     def conv_bson_decimal(self, *args: Any, **kwargs: Any) -> BaseField:
         return DecimalField(**self._standard_type_common(**kwargs))
 
+    @converts("timedelta")
+    def conv_bson_timedelta(self, *args: Any, **kwargs: Any) -> BaseField:
+        return IntervalField(**self._standard_type_common(**kwargs))
+
```
jowilf, line 116: it seems odmantic does not support timedelta https://github.com/art049/odmantic/issues/120, we can remove this and just keep for mysql

regisin:
@jowilf looks like the version of `mkdocstring` needs attention. I did some digging and found this https://github.com/mkdocstrings/autorefs/issues/57

jowilf (approved):
Thank you for the PR, @regisin. I'll look into the mkdocs issue later, as it’s not related to the changes you've made.

jowilf:
Sorry for not merging this; I think it will be difficult to maintain this inside starlette-admin. are you interrested on making this as a plugin https://jowilf.github.io/starlette-admin/advanced/plugins/? There will be a dedicated section in the doc for showcasing plugins

PR #618, opened 2025-01-09
feat(base): add redirect_slashes option to mount_to method

noamsto:
Add optional redirect_slashes parameter to `BaseAdmin.mount_to()` and propagate it to SQLAlchemy Admin class. This allows users to control whether trailing slashes in URLs should be redirected, providing more flexibility in URL handling.

I'm working on a a litestar plugin to this awesome lib.
Inspired by https://github.com/peterschutt/sqladmin-litestar-plugin

Because of Starlette issue: https://github.com/encode/starlette/issues/869
We have to turn of redirect_slashes. In all of the nested Starlette apps.
Because admin_app is created deep in the Base class, it's impossible to do it without hacking the lib.
This PR makes it convenient and is backward compatible.
If you have suggestions how to do it better, let me know!

### DIFF
--- a/starlette_admin/base.py
+++ b/starlette_admin/base.py
@@ -524,7 +524,11 @@ class BaseAdmin:
             data[field.name] = await field.parse_form_data(request, form_data, action)
         return data
 
-    def mount_to(self, app: Starlette) -> None:
+    def mount_to(
+        self,
+        app: Starlette,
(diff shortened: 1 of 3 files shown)


jowilf (changes requested):
thanks for this. please update the mongoengine Admin class as well

jowilf (approved):

Issue #620, opened 2025-01-27
Pre-selection issue in EnumField edit screen when a 0 item is present


tomopy03:
### Description

The edit screen fails to pre-select an `EnumField` value when the underlying `IntEnum` includes an item with the value `0` . This happens because the following code used for determining the pre-selected item treats `0` as a falsy value:

https://github.com/jowilf/starlette-admin/blob/2981324cf469907b608d0740e02add79ab535bdf/starlette_admin/templates/forms/enum.html#L14

When data is 0, the condition `data and ...` evaluates to `False`, preventing the `selected` attribute from being set.

### Steps to Reproduce

Define an `IntEnum` with an item having the value `0` .

```
from enum import IntEnum

class Status(IntEnum):
    INACTIVE = 0
    ACTIVE = 1

class MyModelView(ModelView):
    fields = [EnumField("status", enum=Status)]
```

### Proposed Solution

Modify the condition to explicitly check for `None` instead of relying on the truthiness of `data` . For example:

```
'selected' if (data is not none and (value in (data if field.multiple else [data]))) else ''
```

jowilf:
Nice catch, feel free to submit a PR

tomopy03:
> Nice catch, feel free to submit a PR

Thank you.
Created PR #621 .

PR #621, opened 2025-01-30
Update enum rendering to ensure 'selected' state is applied for data value of 0

tomopy03:
This PR addresses issue #620, where the 'selected' state was not being set correctly for enums when the data value is 0.


jowilf (approved):
thanks for the PR🎉

PR #633, opened 2025-02-28
feat(sqla): add support for joined table polymorphic inheritance

noamsto:
- Add handling for inherited primary keys in SQLAlchemy model converter
- Add test cases for polymorphic inheritance with BaseItem and SpecialItem models
- Update converter to properly handle primary key columns in inherited tables

### DIFF
--- a/starlette_admin/contrib/sqla/converters.py
+++ b/starlette_admin/contrib/sqla/converters.py
@@ -114,14 +114,27 @@ class BaseSQLAModelConverter(BaseModelConverter):
                             )
                         )
                 elif isinstance(attr, ColumnProperty):
-                    assert (
-                        len(attr.columns) == 1
-                    ), "Multiple-column properties are not supported"
-                    column = attr.columns[0]
-                    if not column.foreign_keys:
+                    # Handle inherited primary keys (i.e.: joined table polymorphic inheritance)
+                    is_inherited_pk = mapper.inherits is not None and any(
+                        col.primary_key for col in attr.columns
+                    )
+                    if is_inherited_pk:
+                        column = attr.columns[0]
                         converted_fields.append(
-                            self.convert(name=attr.key, type=column.type, column=column)
+                            self.convert(
+                                name=attr.key, type=column.type, column=column
+                            ),
                         )
+                    else:
+                        assert (
+                            len(attr.columns) == 1
+                        ), "Multiple-column properties are not supported"
+                        column = attr.columns[0]
+                        if not column.foreign_keys:
+                            converted_field = self.convert(
+                                name=attr.key, type=column.type, column=column
+                            )
+                            converted_fields.append(converted_field)
         return converted_fields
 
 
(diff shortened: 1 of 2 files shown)


starlette_admin/contrib/sqla/converters.py
```diff
@@ -114,14 +114,27 @@ def convert_fields_list(
...
+                            ),
                         )
+                    else:
+                        assert (
+                            len(attr.columns) == 1
+                        ), "Multiple-column properties are not supported"
+                        column = attr.columns[0]
+                        if not column.foreign_keys:
+                            converted_field = self.convert(
+                                name=attr.key, type=column.type, column=column
```
jowilf, line 135: I think nothing changes and the unit tests will still pass with the existing code.

jowilf (changes requested):
This was already working with the current code. we can at least keep the unit test

noamsto:
> This was already working with the current code. we can at least keep the unit test

It wasn't working on my other project, I'm gonna recheck.
The differences I can think of is using UUID and not default ID, and using mapped.
I'll report back

noamsto:
```py
class BaseItem(Base):
    __tablename__ = "base_items"
    __mapper_args__ = {
        "polymorphic_identity": "base_item",
        "polymorphic_on": "_type",
    }

    id: Mapped[uuid.UUID] = mapped_column(Integer, primary_key=True)
    name: Mapped[str] = mapped_column(String(50))
    _type: Mapped[str] = mapped_column(String(50), nullable=False)


class SpecialItem(BaseItem):
    __tablename__ = "special_items"
    __mapper_args__ = {"polymorphic_identity": "special_item"}

    id: Mapped[uuid.UUID] = mapped_column(ForeignKey("base_items.id"), primary_key=True)
    special_power: Mapped[str] = mapped_column(String(50))
    
    # Rest of the code
    
    async def test_polymorphic():
    assert ModelView(BaseItem).fields == [
        IntegerField(
            "id", required=True, exclude_from_create=True, exclude_from_edit=True
        ),
        StringField("name", maxlength=50, required=True),
        StringField("_type", maxlength=50, required=True),
    ]
    assert ModelView(SpecialItem).fields == [
        IntegerField(
            "id", required=True, exclude_from_create=True, exclude_from_edit=True
        ),
        StringField("special_power", maxlength=50, required=True),
        StringField("name", maxlength=50, required=True),
        StringField("_type", maxlength=50, required=True),
    ]

```

@jowilf , I found the culprit.
Single table inheritance work as is without any changes.
But Joined table inheritance like the above only works with my additions.

I can restore the changes and modify the test to reflect joined table inheritance.

jowilf:
> @jowilf , I found the culprit.
Single table inheritance work as is without any changes.
But Joined table inheritance like the above only works with my additions.
> 
> I can restore the changes and modify the test to reflect joined table inheritance.

Good catch, probably because of this line https://github.com/jowilf/starlette-admin/blob/9906b53206cf55293b9555d1ca669cdd72460828/starlette_admin/contrib/sqla/converters.py#L121

Feel free to make the changes, thanks

jowilf (approved):
Thank you 🎉

PR #643, opened 2025-03-14
Add `get_details_query` to SQLAlchemy ModelView.

### LINKED ISSUE #642: Enhancement: Allow custom query for details like get_list_query
Currently, Starlette Admin provides `get_list_query`, `get_search_query`, and `get_count_query`, but lacks a `get_details_query` for fetching a single record with a custom query. Adding this would allow more flexibility in retrieving detailed data, similar to how `get_list_query` works.

Would a PR for this be welcome?

nimaxin:
Resolves #642

### DIFF
--- a/starlette_admin/contrib/sqla/view.py
+++ b/starlette_admin/contrib/sqla/view.py
@@ -179,7 +179,8 @@ class ModelView(BaseModelView):
     def get_details_query(self, request: Request) -> Select:
         """
         Return a Select expression which is used as base statement for
-        [find_by_pk][starlette_admin.views.BaseModelView.find_by_pk] method.
+        [find_by_pk][starlette_admin.views.BaseModelView.find_by_pk] and
+        [find_by_pks][starlette_admin.views.BaseModelView.find_by_pks] methods.
 
         Examples:
             ```python  hl_lines="3-4"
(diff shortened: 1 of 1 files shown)


jowilf (changes requested):
Thank you for the PR. I think we need to use it in `find_by_pks` as well

starlette_admin/contrib/sqla/view.py
```diff
@@ -176,6 +176,21 @@ async def handle_row_action(
         except SQLAlchemyError as exc:
             raise ActionFailed(str(exc)) from exc
 
+    def get_details_query(self, request: Request) -> Select:
+        """
+        Return a Select expression which is used as base statement for
+        [find_by_pk][starlette_admin.views.BaseModelView.find_by_pk] method.
```
jowilf, line 182:
```suggestion
        Return a Select expression which is used as base statement for
        [find_by_pk][starlette_admin.views.BaseModelView.find_by_pk] and
        [find_by_pks][starlette_admin.views.BaseModelView.find_by_pks] methods.
```

jowilf (changes requested):

jowilf (approved):

PR #646, opened 2025-03-27
Add zh_Hant (Traditional Chinese) translation

limouren:
It seems like `hatch run i18n:init` didn't handle locale with underscore properly and fell back to `en` for the generated datatables json. I didn't investigate the cause, instead I copied and constructed the json from respective file of https://datatables.net/plug-ins/i18n/.


jowilf (approved):
Thank you

PR #656, opened 2025-04-23
Implement Beanie ODM Support

alexdlukens:
Hello,

I have taken an attempt at adding [Beanie ODM](https://beanie-odm.dev/) support to this project.

I have:
- Converters for Links, List of Links, Pydantic types, and Nested Pydantic models
- Query parsing using mongoengine (filter & sort on Document fields)
- support for overriding search using a "[full text index](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/indexes/#text-indexes)" in MongoDB
- test coverage for my additions

Missing is support for:
- BackLink (I think this is non-trivial, as no document id is stored in the BackLink itself)
  - I render this as a StringField to avoid converter errors

Dependencies needed for my code to run are:
- `beanie`
- `mongoengine`
- `pydantic[email]>2.0` (1.10 not supported)

I note in the older PR on adding Beanie support, using `mongoengine` for query parsing was not preferred. Let me know if this needs to be worked on.

Any feedback is appreciated

Best,


starlette_admin/contrib/beanie/view.py
```diff
@@ -0,0 +1,343 @@
...
+        icon: Optional[str] = None,
+        name: Optional[str] = None,
+        label: Optional[str] = None,
+        identity: Optional[str] = None,
+        converter: Optional[BeanieModelConverter] = None,
+        exclude_fields_from_create: Optional[List[str]] = None,
+        exclude_fields_from_edit: Optional[List[str]] = None,
+        exclude_fields_from_list: Optional[List[str]] = None,
+        exclude_fields_from_detail: Optional[List[str]] = None,
+        full_text_override_order_by: bool = False,
```
jowilf, line 60: Let's remove them from here and keep them as class parameters to be consistent with other backends.

starlette_admin/contrib/beanie/view.py
```diff
@@ -0,0 +1,343 @@
...
+                docs.append(doc)
+        return docs
+
+    async def get_pk_value(self, request: Request, obj: Any) -> Any:
+        if isinstance(obj, Link):
+            return getattr(obj.ref, not_none(self.pk_attr))
+
+        return getattr(obj, not_none(self.pk_attr))
+
+    async def serialize(
```
jowilf, line 235: Was it necessary to override this?

starlette_admin/contrib/beanie/view.py
```diff
@@ -0,0 +1,343 @@
...
+        route_name = request.app.state.ROUTE_NAME
+        obj_meta["detailUrl"] = str(
+            request.url_for(route_name + ":detail", identity=self.identity, pk=pk)
+        )
+        obj_serialized["_meta"] = obj_meta
+        return obj_serialized
+
+    async def create(self, request: Request, data: dict) -> Document:
+        data = {
+            k: v for k, v in data.items() if k not in self.exclude_fields_from_create
```
jowilf, line 301: you should use `self.get_fields_list()` instead

starlette_admin/contrib/beanie/view.py
```diff
@@ -0,0 +1,343 @@
...
+        except ValidationError as ve:
+            raise pydantic_error_to_form_validation_errors(ve) from ve
+        return await doc.create()
+
+    async def edit(
+        self, request: Request, pk: PydanticObjectId, data: dict
+    ) -> Document:
+        doc: Document | None = await self.document.get(pk)
+        assert doc is not None, "Document not found"
+        data = {k: v for k, v in data.items() if k not in self.exclude_fields_from_edit}
```
jowilf, line 314: you should use `self.get_fields_list()` instead

jowilf (changes requested):
Thank you for the PR @alexdlukens. This looks goods. Please review the comment and let’s make sure the test cases succeed.

starlette_admin/contrib/beanie/view.py
alexdlukens: On second thought, I found a better solution instead of overriding serialize. Fixed in 
3c9850e

starlette_admin/contrib/beanie/view.py
alexdlukens: Updated

starlette_admin/contrib/beanie/view.py
alexdlukens: Done 7083157

starlette_admin/contrib/beanie/view.py
alexdlukens: Updated in 7083157

starlette_admin/contrib/beanie/helpers.py
```diff
@@ -0,0 +1,205 @@
...
+    Tuple,
+    Type,
+    Union,
+    get_args,
+    get_origin,
+)
+
+from beanie import Document, Link
+from beanie.odm.enums import SortDirection
+from mongoengine.queryset import Q as BaseQ  # noqa: N811
```
jowilf, line 17: Sorry I missed this earlier. How difficult would it be to decouple the query parser from mongoengine? I'm not very familiar with Beanie, but I was able to build a similar parser in Odmantic without relying on any external dependencies.

starlette_admin/contrib/beanie/view.py
```diff
@@ -0,0 +1,280 @@
...
+        self.name = name or self.name or prettify_class_name(self.document.__name__)
+        self.icon = icon
+        self.pk_attr = "id"
+        self.has_full_text_index: Optional[bool] = None
+
+        self.fields_pydantic = list(document.model_fields.items())
+
+        self.field_infos = []
+        self.link_fields = []
+        self.exclude_fields_from_create = [
```
jowilf, line 75: Can you also add a function that will converts a  list like `[Product.name, Product.description]` into `["name", "description"]` https://github.com/jowilf/starlette-admin/blob/fc995cbbbc797fc37e172829f6c359e81dbbdb0e/starlette_admin/contrib/odmantic/view.py#L67

starlette_admin/contrib/beanie/converters.py
```diff
@@ -0,0 +1,131 @@
...
+            kwargs["name"] = subfield_name
+            kwargs["required"] = subfield_field.is_required()
+            _fields.append(self.convert(*args, **kwargs))
+        return CollectionField(name=name, fields=_fields, required=required)
+
+    def convert_fields_list(
+        self, *, fields: Sequence[Any], model: Type[Any], **kwargs: Dict[str, Any]
+    ) -> Sequence[BaseField]:
+        converted_fields = []
+        for value in fields:
```
jowilf, line 122: this also needs to support ExpressionField. https://github.com/jowilf/starlette-admin/blob/fc995cbbbc797fc37e172829f6c359e81dbbdb0e/starlette_admin/contrib/odmantic/converters.py#L47

alexdlukens: I can do this for sure, but I don't think this makes sense. IIUC you are not supposed to use ExpressionField as a field type. It is just [a subclass of str,](https://github.com/BeanieODM/beanie/blob/157c02d7e8883458c822f0db8325d087731dba6a/beanie/odm/fields.py#L214C7-L214C22) so this would be handled by the default string converter?

jowilf (approved):
Excellent work, @alexdlukens. Thanks for your contribution 🎉

PR #682, opened 2025-06-28
Show and edit datetime in user timezone

### LINKED ISSUE #288: DateTime does not display in browser's timezone, but instead in server timezone, without making that clear
**Describe the bug**
I have dates in the Database mapped like this in SQLALchemy:
```
    created_at = Column(TIMESTAMP(timezone=True, precision=3), nullable=False, server_default=text("CURRENT_TIMESTAMP"))
```

PostgreSQL stores them as ``timestamptz``` (UTC internally).

I want these dates to display in the user's locale / timezone. Unfortunately DateTime is not properly converting things and shows me dates 3h into the future (I am at GMT-03:00). I believe it assumes the dates it reads from the database are the same TZ as the machine running the back-end python code.

The code has to recognize that the DateTime has TZ info, being UTC, and convert to the user's timezone.

**To Reproduce**
Have an entity with the date field like above, stored in the database. Create a new entity with created_at having the time now in utc: ```datetime.datetime.now(datetime.timezone.utc)```.  You will see that the UI shows the DateTime "in the future". It does not print as UTC, so the user thinks it is in local time.  A partial solution would be to properly print "UTC" for these dates. An ideal solution would be to convert to the actual timezone of the browser.

starlette-admin==0.11.1
starlette-admin[i18n]==0.11.1
asyncpg==0.28.0
psycopg2==2.9.6

limouren:
This PR adds `TimezoneConfig`, the ability to:

1. Detect user timezone from browser, and set it on a cookie
2. Read user timezone from such cookie, then display and edit datetime and arrow using user timezone
3. Specify a timezone for database columns that return a native datetime, s.t. datetime from these columns are shown and stored correctly

Fixes #288


limouren:
Fixed the lint error.

limouren:
Okay fixed the lint error again

starlette_admin/fields.py
```diff
@@ -735,21 +737,39 @@ def input_params(self) -> str:
...
 
+        if dt.tzinfo is not None:
+            # Convert timezone-aware datetime to database timezone
+            database_tz = get_database_tzinfo()
+            return dt.astimezone(database_tz).replace(tzinfo=None)
+
+        # dt is naive, assume it's in the user's timezone
+        user_tz = get_tzinfo()
+        database_tz = get_database_tzinfo()
+        return dt.replace(tzinfo=user_tz).astimezone(database_tz).replace(tzinfo=None)
```
jowilf, line 754: This should only happen if the user explicitly sets the timezone config. It will make the transition easier.

starlette_admin/i18n.py
```diff
@@ -163,3 +221,24 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+
+        if self.timezone_config.timezone_cookie_name:
+            cookie_timezone = conn.cookies.get(
+                self.timezone_config.timezone_cookie_name
+            )
+            if cookie_timezone:
+                timezone = cookie_timezone
+
+        set_timezone(timezone or DEFAULT_TIMEZONE)
+        set_database_timezone(self.timezone_config.database_timezone)
```
jowilf, line 243: this the same for every request, i see it as a gobal config than a contextvar

jowilf (changes requested):
Great work on this, and sorry for the late review. I’ve left a few comments.
Also, could we add a dropdown for timezones similar to the one we have for languages, allowing users to change the detected timezone if needed?

starlette_admin/i18n.py
limouren: Yes it is global to a single `Admin`, but a single server might have multiple instances of `Admin`s running. Currently in `parse_form_data` and `serialize_value`, there are no ways to access associated `Admin` instance directly. I think intuitively we could make it available as `Request.admin` with a middleware. What do you think?

jowilf (approved):
Thanks for this @limouren

Issue #696, opened 2025-08-18
Bug: Changes made in before_edit are not saved due to calling replace() on stale validated model


abishek0057:
In the `edit` method, the document is validated and then `replace()` is called on the validated copy **before** running `before_edit`. This causes any changes made in `before_edit` to be lost.

#### Current code:
```python
#starlette_admin/contrib/beanie/view.py
 async def edit(self, request: Request, pk: PydanticObjectId, data: dict) -> Any:
  ...
  validated_doc: T = self.document.model_validate(doc.model_dump())
  await self.before_edit(request, data=data, obj=doc)
  updated_doc = await validated_doc.replace()
  ...
```

#### Problem:
replace() is called on validated_doc, which was created before before_edit. So if before_edit modifies doc, those changes are not reflected in the saved document.

**Environment:**
 - Starlette-Admin version: [e.g. 0.15.1]
 - ORM/ODMs: [Beanie ODM]

PR #697, opened 2025-08-18
fix(beanie): move validation after before_edit to persist hook change…

abishek0057:
…s (#696)


abishek0057:
@jowilf I have pushed the suggested code. Please have a look

jowilf (approved):
good, thanks for the fix

PR #707, opened 2025-09-16
Fix one-to-many relation create/edit with SQLAlchemy

### LINKED ISSUE #485: Bug: Required foreign key is empty
**Describe the bug**
When creating a new entity which has required related model I've got an error.

**To Reproduce**
```
class Company(AddressMixin, TimeMixin, table=True):
    id: int = Field(default=None, primary_key=True)
    name: str 
    facilities: List["Facility"] = Relationship(back_populates="company")

class Facility(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    name: str
    company_id: int = Field(foreign_key="company.id")
    company: Company = Relationship(back_populates="facilities")
```

During saving the Facility there is an 422 http error and "{'company_id': 'Field required'}" in errors.

 - Starlette-Admin version: starlette-admin = {extras = ["i18n"], version = "^0.13.0"}
 - ORM/ODMs: SQLModel sqlmodel = "^0.0.14"

tmshlvck:
One-to-many relations do not automatically fill the relation_id foreign key, which causes 422 errors propagation with no save to the DB

By this patch we bind the foreign key column with the relation "virutal" column in SQLModel / SQLAlchemy - the name of the foreign key has to be "<relation>_id", which corresponds with the existing documentation.

This closes https://github.com/jowilf/starlette-admin/issues/485 and probably two related/duplicate issues.

### DIFF
--- a/starlette_admin/contrib/sqla/view.py
+++ b/starlette_admin/contrib/sqla/view.py
@@ -543,6 +543,7 @@ class ModelView(BaseModelView):
                     arranged_data[field.name] = await foreign_model.find_by_pk(
                         request, data[field.name]
                     )
+                    arranged_data[f"{field.name}_id"] = arranged_data[field.name].id
             else:
                 arranged_data[field.name] = data[field.name]
         return arranged_data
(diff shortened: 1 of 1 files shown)


jowilf:
This fix is not generalized. fixed in #798

Issue #706, opened 2025-09-16
Bug: Numerical filter with 0 is sent to backend as ""


MajorXaker:
**Describe the bug**
When you type in a numerical value 0 to filters, it is being sent to backend as empty string "". Later on when sqlalchemy tries to compare numerical value in database to the empty string, the exception is raised: `sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type integer: ""`

<img width="694" height="261" alt="Image" src="https://github.com/user-attachments/assets/67948cec-ea06-4499-968b-41d5bf9bd816" />

is sent as `http://127.0.0.1:8000/admin/api/numbers?skip=0&limit=20&order_by=id desc&where={"and":[{"value":{"gt":""}}]}`

**To Reproduce**
To reproduce you'd need a PostgreSQL database you can connect to. 
Here is complete python script that could show this issue:

```
from fastapi import FastAPI
from sqlalchemy import Column, Integer, create_engine, text
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import sessionmaker
from starlette_admin.contrib.sqla import Admin, ModelView
from starlette.middleware.sessions import SessionMiddleware
import uvicorn

DATABASE_URL = f"postgresql://postgres:postgres@localhost:5432"
DB_NAME = "starlette_poc"


engine = create_engine(f"{DATABASE_URL}/{DB_NAME}")
SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)

Base = declarative_base()


class Numbers(Base):
    __tablename__ = "numbers"

    id = Column(Integer, primary_key=True, index=True)
    value = Column(Integer, nullable=False)

    def __str__(self):
        return f"Number #{self.id}: {self.value}"


def create_database_if_not_exists():
    """Create the database if it doesn't exist"""
    admin_url = f"{DATABASE_URL}/postgres"
    admin_engine = create_engine(admin_url)

    try:
        with admin_engine.connect() as conn:
            # Check if database exists
            result = conn.execute(text(f"SELECT 1 FROM pg_database WHERE datname = '{DB_NAME}'"))
            exists = result.fetchone()

            if not exists:
                conn.execute(text("COMMIT"))
                conn.execute(text(f"CREATE DATABASE {DB_NAME}"))
                print(f"Database '{DB_NAME}' created successfully")
            else:
                print(f"Database '{DB_NAME}' already exists")

    except Exception as e:
[...]

PR #708, opened 2025-09-16
Fix umerical filter with 0 is sent to backend as ""

MajorXaker:
Fixes issue #706 

I've changed `||` to `??` so 0 would be sent to BE as well as `false` which might be a text somewhere (though probably it's less common).


MajorXaker:
Since bitnami closed access to their new repos, I've switched here to "bitnamilegacy" docker repository.

PR #715, opened 2025-11-12
Fix CustomView is_active not working for custom root path

Alwinator:
My custom view "docs" was not highlighted in the navigation because the path `/admin/docs` was compared to `/docs`. This is fixed my this PR.

Example:
```python
admin = Admin(engine, title="MyAdminView")
admin.add_view(
    CustomView(
        label="API Docs",
        path="/docs",
        template_path="docs.html",
        icon="fa fa-book"
    )
)
admin.mount_to(app)
```

### DIFF
--- a/starlette_admin/views.py
+++ b/starlette_admin/views.py
@@ -171,7 +171,7 @@ class CustomView(BaseView):
         )
 
     def is_active(self, request: Request) -> bool:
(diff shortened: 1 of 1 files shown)


Alwinator:
@jowilf Would you be so kind and could check? :)

PR #740, opened 2026-03-03
added optional jsonschema validation to jsonfield

### LINKED ISSUE #735: Enhancement: Add optional jsonschema validation to JSONField
I would like to be able to have my JSON fields respect a predefined schema and have this be validated client-side.

Since [jsoneditor already has builtin support for schema validation](https://github.com/josdejong/jsoneditor/blob/master/docs/api.md#configuration-options), it would be nice if this can be exposed in starlette-admin.

It seems a possible implementation would not require a big change, it is mostly a matter of passing in the 
additional `schema` configuration option when creating the jsoneditor. This could be done using a `data-schema` attribute, similarly to the already existing `data-modes` one.

The `fields.JSONField` class could then have an additional optional `schema` attribute, which would be the schema.

If you find this a valuable addition I can likely contribute a PR with it, for further review

ricardogsilva:
This PR adds optional JSONSchema validation to `JSONField` by leveraging the already existing support for this in JSONEditor.

It introduces a new `validation_schema` attribute to `JSONField`, which can optionally be used to pass a dict with the JSON schema. For example:

```python


class TodoView(ModelView):
    fields = (
        Todo.title,
        Todo.done,
        JSONField(
            "stuff",
            validation_schema={
                "type": "object",
                "required": ["owner"],
                "properties": {
                    "owner": {"type": "string"},
                }
            }
        ),
    )
```

When using the admin to create or edit an instance, the UI will show a small symbol if there are validation errors with a descriptive tooltip message, as seen below:

image

This PR really only glues things up for starlette-admin to use the default functionality provided in upstream JSONEditor - it does not add any custom validation event handlers or any type of further customization.

---
- fixes #735

### DIFF
--- a/starlette_admin/fields.py
+++ b/starlette_admin/fields.py
@@ -936,10 +936,15 @@ class ArrowField(DateTimeField):
 class JSONField(BaseField):
     """
     This field render jsoneditor and represent a value that stores python dict object.
-    Erroneous input is ignored and will not be accepted as a value."""
+    Erroneous input is ignored and will not be accepted as a value.
+
+    You may optionally use the `validation_schema` property to provide a dict with
+    a [JSONSchema](https://json-schema.org/) to have it be used to provide
+    validation feedback on the client side."""
 
     height: str = "20em"
     modes: Optional[Sequence[str]] = None
+    validation_schema: Optional[Dict[str, Any]] = None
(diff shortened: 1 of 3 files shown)


jowilf (approved):
thanks for your contribution @ricardogsilva

PR #741, opened 2026-03-04
Quote jsonfield name when selecting in order to account for possible numeric parts in the name

### LINKED ISSUE #737: Bug: Cannot have multiple JSONField in a ListField
It is currently not possible to have multiple JSONFields in a ListField, as this causes a JS syntax error.

Take this model view:

```python
from starlette_admin.contrib.sqlmodel impot ModelView
from starlette_admin.fields import JSONField, ListField

class SomeModelView(ModelView):
    fields = (
        ListField(
            JSONField(name="sad_title"),
        ),
    )
```

When a user accesses the admin and clicks the '➕ Add item' button, a JS error appears:

```javascript
Uncaught Error: Syntax error, unrecognized expression: input[name=sad_title.1]
```

This error is coming from

https://github.com/jowilf/starlette-admin/blob/4816ae594679ae3f155b30a27323cb6e849ddb30/starlette_admin/statics/js/form.js#L36-L49

Seems like a possible fix for this would be to simply quote the value of `name`:

```diff
$("div.field-json", element).each(function () {
    let el = $(this);
    let name = el.attr("id");
    new JSONEditor(
      this,
      {
        modes: String(el.data("modes")).split(","),
        onChangeText: function (json) {
-          $(`input[name=${name}]`).val(json);
+          $(`input[name='${name}']`).val(json);
        },
      },
-      JSON.parse($(`input[name=${name}]`).val())
+      JSON.parse($(`input[name='${name}']`).val())
    );
  });
```

When I add my own modified copy of `form.js` with this change the UI correctly renders the multiple JSONFields, which makes it possible to work with JSONField and ListField together.

I can prepare a PR for this if you would be interested in the change

ricardogsilva:
This PR makes a small change to the way JSONFieldfields are selected during their respective field initializer function. The change is merely quoting the element's name when finding it with JQuery.

As mentioned in #737, this changes makes it possible to use JSONField together with ListField.

---

- fixes #737

### DIFF
--- a/starlette_admin/statics/js/form.js
+++ b/starlette_admin/statics/js/form.js
@@ -42,10 +42,10 @@ registerFieldInitializer(function (element) {
         modes: String(el.data("modes")).split(","),
         schema: el.data("validationSchema") ?? undefined,
         onChangeText: function (json) {
-          $(`input[name=${name}]`).val(json);
+          $(`input[name='${name}']`).val(json);
         },
       },
-      JSON.parse($(`input[name=${name}]`).val())
+      JSON.parse($(`input[name='${name}']`).val())
     );
   });
 


jowilf (approved):
Thanks for the fix

PR #742, opened 2026-03-04
Fix docs on field customization

### LINKED ISSUE #736: Bug: custom field docs don't match code implementation
The ['Data processing' section of the docs on Custom Field](https://jowilf.github.io/starlette-admin/advanced/custom-field/#data-processing) needs to be updated. It mentions a `process_form_data()` function in the text - this function does not exist. 

The code example provided in the docs is also not correct, as the signature for `parse_form_data()` includes the additional `action` parameter.

ricardogsilva:
This PR fixes the docs section that describes field customization by correcting the erroneous code samples

---

- fixes #736

### DIFF
--- a/docs/advanced/custom-field/index.md
+++ b/docs/advanced/custom-field/index.md
@@ -132,7 +132,7 @@ class CustomField(BaseField):
 
 For data processing you will need to override two functions:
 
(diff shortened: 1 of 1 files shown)


jowilf (approved):
Thanks @ricardogsilva

PR #745, opened 2026-04-02
Add Foreign Key Filter Support in SearchBuilder

### LINKED ISSUE #717: Enhancement: Foreign Key filter
How to filter by Foreign Key? I have 2 tables, Roles and Users (each user has a foreign key to the Roles table). The filters don't filter, but I need to display all users with a certain role.

Alwinator:
## Related Issues
- Closes #717

## What this does

This PR adds the ability to filter by foreign key (relation) fields directly in the DataTables SearchBuilder UI. Previously, relation fields (`HasOne` / `HasMany`) only supported `null` / `!null` conditions in the SearchBuilder. Now they behave like string columns and support the full set of string conditions: `=`, `!=`, `contains`, `starts`, `ends`, and their negated counterparts.

When a user types a search term (e.g. "John") for a relation column, the backend resolves it by searching across all string columns of the related model using case-insensitive matching. This effectively matches against what `__admin_repr__` would show — so the filter feels intuitive from the user's perspective.

## How it works

- The SearchBuilder treats relation fields as `string` type columns, so users get a free-text input (no dropdown, no extra AJAX calls).
- On the backend, when a string operator hits a relationship attribute, we introspect the related model's string columns and build `.has()` (for `HasOne`) or `.any()` (for `HasMany`) clauses with `OR`'d `ilike` conditions.
- Negated operators (`not_contains`, `neq`, etc.) use `AND` logic so that a row is only excluded if *none* of the related model's string columns match.
- `is_null` / `is_not_null` continue to work exactly as before.

No frontend JS changes were needed — the existing `string` SearchBuilder type and `extractCriteria()` handle everything already since we're in `serverSide: true` mode.

## Performance escape hatch

For large datasets where searching all string columns on a related model might be expensive, there's a new `searchable_relation_fields` option:

```python
class PostView(ModelView):
    searchable_relation_fields = {
        "publisher": ["last_name", "first_name"],
    }
```

This restricts the generated query to only the specified columns instead of scanning every string column on the related model. By default (when not set), all string columns are searched.

## Changes

- **`starlette_admin/fields.py`** — `RelationField.search_builder_type` changed from `"default"` to `"string"`
- **`starlette_admin/views.py`** — Added `searchable_relation_fields` attribute to `BaseModelView`
- **`starlette_admin/contrib/sqla/view.py`** — Relation fields are now included in the default `searchable_fields` list (but still excluded from `sortable_fields`); `searchable_relation_fields` is threaded through to the query builder
- **`starlette_admin/contrib/sqla/helpers.py`** — New `_relationship_string_filter()` and `_apply_string_operator()` helpers; `build_query()` now detects string operators on relationship attributes and delegates accordingly
- **`tests/sqla/test_view_meta.py`** — Updated assertions to reflect that relation fields are now searchable by default
- **`tests/sqla/test_fk_filter.py`** — 28 new tests covering HasOne filters, HasMany filters, AND/OR combinations, the `searchable_relation_fields` allowlist, reverse relations, case-insensitive matching, and multi-column search

## Scope

This PR only implements FK filtering for the **SQLAlchemy backend** (including SQLModel, which inherits it automatically). Support for ODMantic, MongoEngine, and Beanie is deferred to a follow-up — their converters would need similar treatment but the patterns differ enough to warrant separate work.

## Testing

All existing SQLA tests continue to pass. The 28 new tests cover:

- Every string operator on `HasOne` relations (contains, eq, neq, startswith, endswith, and all negations)
- `HasMany` relation filtering (uses `.any()` instead of `.has()`)
- `is_null` / `is_not_null` (unchanged behavior, but verified alongside the new operators)
- AND/OR combinations mixing FK filters with regular field filters
- The `searchable_relation_fields` allowlist (verifying that excluded columns are truly not searched)
- Reverse relation filtering (e.g. filtering Publishers by their Books)
- Case-insensitive matching
- Multi-column search (matching against *any* string column on the related model)

### DIFF
--- a/starlette_admin/contrib/sqla/helpers.py
+++ b/starlette_admin/contrib/sqla/helpers.py
@@ -1,6 +1,7 @@
-from typing import Any, Callable, Dict, Optional, Sequence
+from typing import Any, Callable, Dict, List, Optional, Sequence
 
 from sqlalchemy import String, and_, cast, false, not_, or_, true
+from sqlalchemy import inspect as sa_inspect
 from sqlalchemy.orm import (
     InstrumentedAttribute,
     RelationshipProperty,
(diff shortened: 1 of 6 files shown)


jowilf:
Thanks @Alwinator  for the PR, i like the proposition. I will review the code and merge if it looks good

jowilf:
datatables has been dropped

Issue #515, opened 2024-02-29
Bug: Filter on Enum field does not constrain values


sglebs:
**Describe the bug**

An enum field should only allow me to set it to one of the enum possible values. But the UI is "free text"

<img width="730" alt="image" src="https://github.com/jowilf/starlette-admin/assets/1507221/6a535242-367c-41ce-bcc6-a4b312744036">

**To Reproduce**

Just add an enum field to your model, with 4 pre-defined values, and try to use a filter in the UI.

**Environment (please complete the following information):**

- Starlette-Admin version: 0.13.1
- ORM/ODMs: SQLAlchemy, psycopg2-binary==2.9.9
- SQLAlchemy-serializer==1.4.1

**Additional context**

None

nimaxin:
Any updates or plans to address this issue?

jowilf:
Fixed in #746

PR #746, opened 2026-04-14
Replace input with select for EnumField filter (#515)

shana0440:
This PR updates the filter UI by replacing the text input with a select dropdown for the EnumField, providing a better UX.

image

Related issues: 
#515


jowilf (approved):
Thanks for the PR @shana0440

PR #441, opened 2023-12-27
Support translation for delete row action submit button text.

hasansezertasan:
I didn't do anything with the i18n commands like `compile` or `extract` because those were removed at #425.

PR #752, opened 2026-04-22
Make single-row delete submit button translatable (closes #751)

### LINKED ISSUE #751: "Yes, delete" button in row delete confirmation is not translatable
**Describe the bug**  
The built-in single-row delete action uses a hardcoded English submit button label: `Yes, delete`.

This string is not wrapped in `_()` and therefore is not passed through the translation system.

https://github.com/jowilf/starlette-admin/blob/0f684fe82964484108d123dca17e4644377dc0d6/starlette_admin/views.py#L503

**To Reproduce**  
1. Use starlette-admin with a non-English locale  
2. Trigger the single-row delete action  
3. Observe that the confirmation button label remains `Yes, delete` instead of being translated  

**Suggested fix**

Update the code to: `submit_btn_text=_("Yes, delete"),`

sjhddh:
## Summary
- Wrap the single-row delete action's `submit_btn_text="Yes, delete"` with `_()` so it honors the active locale. The neighboring `text=_("Delete")` and `confirmation=_(...)` on the same `@row_action` were already translated; this literal was the only hold-out.
- Regenerated `i18n/admin.pot` and every `admin.po`/`admin.mo` via `hatch run i18n:{extract,update,compile}`.
- Added translations for de / fr / pt / ru / tr / zh_Hant (reused the "Yes, delete all" pattern, just without the "all" quantifier). Left `en` msgstr empty per existing convention.

Repro:

```python
from starlette_admin.i18n import set_locale, gettext
set_locale("fr"); gettext("Yes, delete")  # before: 'Yes, delete' — after: 'Oui, supprimer'
```

## Test plan
- [x] `hatch run test:test tests/test_i18n.py` — 16/16 pass locally
- [x] Runtime spot-check: loaded each supported locale and called `gettext("Yes, delete")` — all return the localized string
- [x] Single-line source change is line-for-line identical to the closed #441 approach the maintainer previously endorsed; catalog churn is pure `pybabel extract/update/compile` output (line-ref and POT-Creation-Date bumps)

## Notes
This is the same fix as the closed #441 by @hasansezertasan (closed by the author due to inactivity, not on technical grounds — the maintainer's own review comment confirmed the translation was desired). Reopening here because #751 reported the bug is still live in main as of 0f684fe.


sjhddh:
Heads-up on the red pre-commit.ci: the failure is unrelated to this PR. `pyupgrade v3.19.1` (pinned in `.pre-commit-config.yaml`) crashes on Python 3.14.1+ with `TypeError: cannot use a bytes pattern on a string-like object` in `tokenize.cookie_re.match` — pre-commit.ci's runner appears to be on Python 3.14.1 now, and the last main CI run (2026-04-15) was on an older 3.14.

Upstream fix landed in asottile/pyupgrade#1040 (released as v3.20.0+). Running the same hooks locally on Python 3.9 against this branch passes clean. Happy to open a separate tiny PR bumping the pyupgrade rev in `.pre-commit-config.yaml` if that's useful — didn't want to expand the scope of this one.

jowilf:
fixed in #798

PR #753, opened 2026-04-22
Fix untranslated "Yes, delete" button in row delete confirmation

### LINKED ISSUE #751: "Yes, delete" button in row delete confirmation is not translatable

Markkuuss:
Fixes #751

Wrap the `Yes, delete` label in `_()` so it is passed through the translation system.

Update `.po` files and regenerate `.mo` files.


jowilf:
fixed in #798

PR #756, opened 2026-04-30
Add `hidden_from_list` field attribute

Alwinator:
### Summary

Adds a new `hidden_from_list` attribute to `BaseField` that allows hiding columns from the list view by default while still keeping them accessible via DataTables' Column Visibility button.

### Motivation

Currently, the only way to remove a column from the list view is `exclude_from_list`, which completely removes the column from the DataTable — it cannot be toggled back by the user, and it is no longer available for filtering or searching.

There are cases where a field contains useful but non-essential information (e.g. long descriptions, internal IDs, timestamps) that should be hidden by default to reduce clutter, but still available for users who want to see it. Additionally, hidden fields can still be used for filtering and searching via Search Builder, which is valuable when users need to query on columns they don't need to see in every row.

`hidden_from_list` fills this gap: the column is rendered in the DataTable with `visible: false` via `columnDefs`, so users can reveal it using the Column Visibility button (enabled by default via `column_visibility = True`), and the column remains available for filtering and sorting.

### Changes

**`starlette_admin/fields.py`**
- Added `hidden_from_list: Optional[bool] = False` attribute to `BaseField` dataclass.
- Added documentation for the new parameter.

**`starlette_admin/views.py`**
- Replaced direct pass-through of `self.datatables_options` in `_configs()` with a call to the new `_merged_datatables_options()` method.
- Added `_merged_datatables_options(request)` which deep-copies user-defined `datatables_options`, detects fields with `hidden_from_list=True`, and appends a `columnDefs` entry with `{"visible": False, "targets": [...]}` (offset by +1 to account for the leading checkbox column).

**`tests/test_base_model_view.py`**
- `test_hidden_from_list_field_attribute` — verifies the attribute defaults to `False` and can be set to `True`.
- `test_hidden_from_list_still_in_field_list` — confirms hidden fields are still included in `get_fields_list` (unlike `exclude_from_list`).
- `test_hidden_from_list_excluded_field_not_in_list` — confirms that `exclude_from_list=True` takes precedence and removes the field entirely.
- `test_merged_datatables_options_hidden_columns` — verifies correct `columnDefs` targets are generated for multiple hidden fields.
- `test_merged_datatables_options_no_hidden` — verifies no `columnDefs` are injected when there are no hidden fields.
- `test_merged_datatables_options_preserves_user_options` — verifies existing `datatables_options` keys are preserved.
- `test_merged_datatables_options_merges_existing_column_defs` — verifies the hidden column def is appended (not overwritten) when the user already has `columnDefs`.

### Usage

```python
class ArticleView(ModelView):
    fields = [
        IntegerField("id"),
        StringField("title"),
        TextAreaField("body", hidden_from_list=True),  # hidden by default, toggleable, filterable
        DateTimeField("created_at", hidden_from_list=True),
    ]
```

### Backwards Compatibility

Fully backwards-compatible. `hidden_from_list` defaults to `False`, so existing views are unaffected. User-defined `datatables_options` (including existing `columnDefs`) are preserved and merged.


jowilf:
datatables has been dropped

Issue #734, opened 2026-02-02
Bug: `JsonField` incorrectly renders falsy values as null or empty objects


nimaxin:
**Describe the bug**
`JsonField` does not correctly display falsy values. Values such as `0`, `false`, `""`, `[]`, and `{}` are rendered as `-null-` in the list view and as `{}` in form views, even though the actual stored values are different.

**To Reproduce**
1 - Create a `JsonField`.
2 - Store any of the following values: `0`, `false`, `""`, `[]`, or `{}`,
3 - Open the list view and form view for the model.

**Environment (please complete the following information):**
 - Starlette-Admin version: 0.16.0
 - ORM/ODMs: SQLAlchemy

**Additional context**
This behavior is misleading when working with JSON fields that legitimately store falsy values, as it visually conflates them with null and can lead to incorrect assumptions or data edits.

jowilf:
could not reproduce with `starlette-admin==1.0.0rc3`

PR #765, opened 2026-05-17
fix: JsonField now correctly renders falsy values in list view (#734)

SHYXIN:
## Summary
The json render function used `if (data)` which incorrectly treated falsy values (`0`, `false`, `""`, `[]`, `{}`) as null, displaying them as `-null-` in the list view.

## Changes
- `starlette_admin/statics/js/render.js:67` — Changed `if (data)` to `if (data != null)` so only actual null/undefined displays as `-null-`
- `tests/test_views.py` — Added `test_json_field_falsy_values` to verify API correctly serializes falsy JSON values

## Testing
- New test passes: `test_json_field_falsy_values`
- All 15 views tests pass

### DIFF
--- a/starlette_admin/statics/js/render.js
+++ b/starlette_admin/statics/js/render.js
@@ -64,7 +64,7 @@ const render = {
   },
   json: function render(data, type, full, meta, fieldOptions) {
     if (type != "display") return escape(JSON.stringify(data));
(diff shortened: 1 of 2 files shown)


jowilf:
datatables is dropped

Issue #732, opened 2026-01-21
Bug: TextArea removes first blank line on edit/save


sglebs:
**Describe the bug**
I have a text area field that is supposed to be multi-line, each one with important text and "empty" is important. When I edit the field and leave the first line empty/blank and "Save and Continue Editing", it comes back without the first, blank line. It is as if a strip() call is being used.

**To Reproduce**
Add a database field:

human_support_business_hours = Column(Text)

And its corresponding View.  Save it containing, for example, 5 different lines:

<img width="333" height="140" alt="Image" src="https://github.com/user-attachments/assets/4dfc4aaf-8122-4158-9cc2-f07a78278503" />

Now make the first line blank. Save and continue editing. The bank line disappears.

**Environment (please complete the following information):**
  * Starlette-Admin version: 0.16.0
  * ORM/ODMs: SQLAlchemy, psycopg2-binary==2.9.9
  * SQLAlchemy-serializer==1.4.1

**Additional context**
Leading blanking lines is important in my case, and how many as well. I don't want this side-effect.

jowilf:
fixed in #798 (can be tested with starlette-admin==1.0.0rc4)

PR #767, opened 2026-05-17
fix: preserve leading newlines in TextArea field (#732)

SHYXIN:
## Summary
TextArea field values with leading newlines (e.g. `\n\nline2`) lost the first newline when rendered in the edit form. This is because the HTML spec states that browsers ignore the first newline immediately after the `<textarea>` opening tag.

## Root Cause
The original template used `{% if data -%}` with whitespace-trimming (`-%}`), which produced:
```
<textarea...>{{ data }}</textarea>
```
When `data` was `\n\nline2`, the output was `<textarea>\n\nline2</textarea>`, and the browser consumed the first `\n`.

## Fix
`starlette_admin/templates/forms/textarea.html` — Removed whitespace-trimming and added an explicit newline before `{{ data }}`:
```
<textarea...>
{{ data }}</textarea>
```
Now the browser consumes the template's newline, preserving the data's leading newlines.

## Testing
- Verified API returns correct value with leading newlines
- Verified edit page textarea content preserves leading newlines
- Verified "Save and Continue Editing" cycle preserves content


jowilf:
fixed in #798

Issue #486, opened 2024-01-23
Bug: required relationship field is not marked as required


mlachowski:
**Describe the bug**

Required relationship field is not marked as required (as for example "Name" is marked).

![image](https://github.com/jowilf/starlette-admin/assets/49247459/4d0ed146-5bbc-4185-9833-2d7271314110)

**To Reproduce**
Create simple model and view with following fields:
```
class Company(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    name: str 
    facilities: List["Facility"] = Relationship(back_populates="company")

class Facility(SQLModel, table=True):
    id: int = Field(default=None, primary_key=True)
    name: str
    company_id: int = Field(foreign_key="company.id")
    company: Company = Relationship(back_populates="facilities")

company_view = ModelView(Company)
facility_view = ModelView(Facility)
```

**Environment (please complete the following information):**
 - Starlette-Admin version: starlette-admin = {extras = ["i18n"], version = "^0.13.0"}
 - ORM/ODMs: sqlmodel = "^0.0.14"

sglebs:
I don't get any of these mandatory markers, even for simple fields. But I use SQLAlchemy. Is it possible to have it?

IDmedia:
I'm trying to switch to starlette-admin and have noticed the same. Any good solution?

I've tried adding this on my user relationship, but then the edit-form gives me a 404 error:
```
fields = [
        RelationField("user", required=True),
    ]
```

regisin:
Just noticed the issue.. any update?

regisin:
@jowilf I was diggin into this, could you tell me if I'm on the right track?

1. Change converter to set required on HasOne object:
  
    In `starlette_admin\contrib\sqla\converters.py`, method `convert_fields_list(...)`, line 107:
    
    Was: ` converted_fields.append(HasOne(attr.key, identity=identity))`
    
    Update to:

    ```python
    converted_fields.append(
                                HasOne(
                                    attr.key,
                                    required=not any(col.nullable for col in attr.local_columns),
                                    identity=identity
                                )
                            )
    ```
    
    This change makes the red * show up in the form.

2. Update the form field:
    
    In `starlette_admin\templates\forms\relation.html`, add prop for required fields:
    
    was:
    ```html
    ...
    <select data-allow-clear="true"
                ...
                {% if field.multiple %}multiple{% endif %}>
        </select>
    ...
    ```
    
    Update to:
    
    ```html
    ...
[...]

PR #768, opened 2026-05-17
fix: mark HasOne as required when foreign key is NOT NULL (#486)

SHYXIN:
## Summary
When a MANYTOONE relationship's local foreign key column is NOT NULL, the HasOne field was not marked as required in forms. This caused the form to not show the required indicator and not enforce the constraint.

## Root Cause
In `BaseSQLAModelConverter.convert_fields_list`, `HasOne` was always created with `required=False` regardless of the foreign key column's nullable setting.

## Fix
`starlette_admin/contrib/sqla/converters.py:107-113` — Check all local columns of the RelationshipProperty. If all local columns are NOT NULL, set `required=True` on the HasOne field.

```python
required = all(not col.nullable for col in attr.local_columns)
converted_fields.append(HasOne(attr.key, identity=identity, required=required))
```

## Testing
- `tests/test_converters.py` — 2 tests pass
- `tests/test_views.py` — 14 tests pass
- Manual verification with nullable and non-nullable FK columns

## Additional
Includes a chore commit to fix pyproject.toml pytest config conflict that was blocking test execution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,9 +183,6 @@ parallel = true
 concurrency = ["thread", "greenlet"]
 source = ["starlette_admin", "tests"]
 
-[tool.pytest]
-asyncio_mode = "auto"
-
 [tool.pytest.ini_options]
 asyncio_mode="auto"
 asyncio_default_fixture_loop_scope="function"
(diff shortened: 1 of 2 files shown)


jowilf:
was fixed in #798 https://github.com/jowilf/starlette-admin/blame/9002bdc37cdd6c494c61ec363b2cb56f73aa063b/starlette_admin/contrib/sqla/converters.py#L228

Issue #681, opened 2025-06-27
Enhancement: JSONField should allow for pretty printing


fh-igor-serko:
**Is your feature request related to a problem? Please describe.**
When using the JSONField I would want to see the JSON content pretty printed.

**Describe the solution you'd like**
Currently `starlette-admin` uses the [json.html template](https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/templates/displays/json.html) for rendering JSONField.
It uses the `tojson` filter, initial assumption was that it's the [Jinja2's version](https://github.com/pallets/jinja/blob/main/src/jinja2/filters.py#L1690-L1715), but it is not and [starlette-admin seems to override it](https://github.com/jowilf/starlette-admin/blob/main/starlette_admin/base.py#L241).

So I think the solution here is to just remove the override and rely on Jinja2's internal version. Not sure whether `default=str` is something we need to worry about though.

The solution for pretty printing would then be to create a separate template file in your application, like:

```
# templates/admin/json_pretty_print.html
<div id="{{ field.name }}" class="field-json">{{ data|tojson(indent=2) }}</div>
```

And then use it on a JSONField, like:

```
class MyModelView(ModelView):
    fields = [
        JSONField("json_data", display_template="admin/json_pretty_print.html")
    ]
```

**Describe alternatives you've considered**
- Could just add the indent option to the lambda function. However looking closely at the Jinja implementation, it does run htmlsafe over it, while the starlette-admin lambda doesn't.
- I found no way to override which filters get loaded into `starlette-admin`

**Additional context**
n/a

jowilf:
in the upcoming release JSON field will render with a jsonviewer. If you want a simple json text viewer you could overriede the template in your project. just provide your own fields/list/json.html and fields/detail/json.html

PR #769, opened 2026-05-17
feat: support pretty printing in tojson filter (#681)

SHYXIN:
## Summary
The custom `tojson` filter now accepts an optional `indent` parameter, enabling pretty printed JSON output in templates. Default behavior is unchanged.

## Changes
- `starlette_admin/base.py:255` — Added `indent=None` parameter to the `tojson` filter lambda

## Usage
```jinja2
{{ data|tojson }}           # compact (default, backward compatible)
{{ data|tojson(indent=2) }} # pretty printed with 2-space indent
```

Users can override the JSONField display template to use pretty printing:
```python
JSONField("json_data", display_template="admin/json_pretty_print.html")
```

With a custom template:
```html
<div class="field-json">{{ data|tojson(indent=2) }}</div>
```

## Testing
- All 16 tests pass (test_views.py + test_converters.py)

## Additional
Includes chore commit to fix pyproject.toml pytest config conflict.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,9 +183,6 @@ parallel = true
 concurrency = ["thread", "greenlet"]
 source = ["starlette_admin", "tests"]
 
-[tool.pytest]
-asyncio_mode = "auto"
-
 [tool.pytest.ini_options]
 asyncio_mode="auto"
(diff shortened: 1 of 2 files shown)


jowilf:
jsonfield now uses JSonviewer to render

Issue #690, opened 2025-07-23
Auto-prefill form fields from SQLAlchemy model defaults


LoveBloodAndDiamonds:
Is your feature request related to a problem? Please describe.
Currently, when creating a new record via Starlette Admin, all form fields are empty by default, even if the SQLAlchemy model defines default values (either through default= or server_default=). This means users have to manually fill in values that should already have sensible defaults.

⸻

Describe the solution you’d like
I would like Starlette Admin to automatically prefill form fields with the default values defined in the SQLAlchemy model.
For example, when a user clicks “Create” for a model, the form should already display those default values, similar to how sqladmin handles this.

⸻

Describe alternatives you’ve considered
	•	Manually overriding form fields and setting initial values in a custom view.

jowilf:
fixed in #798

PR #771, opened 2026-05-17
feat: pre-fill create form with SQLAlchemy column defaults (#690)

SHYXIN:
## Summary
When rendering the create form, extract default values from the SQLAlchemy model's column definitions and pass them as the initial `obj` to the template. This pre-fills form fields with their database-defined defaults.

## Changes
- `starlette_admin/base.py` — Added `_get_default_values()` method that reads `column.default.arg` from the model's `__table__`
- `_render_create()` — On GET requests, calls `_get_default_values()` and passes the result as `obj` to the template context

## How it works
For SQLAlchemy models, the method iterates over `model.__table__.columns` and extracts `col.default.arg` for columns that have a Python-side default. For non-SQLAlchemy models, returns an empty dict (no change in behavior).

Example: a column defined as `status = Column(String(20), default='active')` will pre-fill the form field with `'active'`.

## Testing
- All 14 views tests pass
- Manual verification with SQLAlchemy model defaults

## Additional
Includes chore commit to fix pyproject.toml pytest config conflict.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,9 +183,6 @@ parallel = true
 concurrency = ["thread", "greenlet"]
 source = ["starlette_admin", "tests"]
 
-[tool.pytest]
-asyncio_mode = "auto"
-
 [tool.pytest.ini_options]
 asyncio_mode="auto"
 asyncio_default_fixture_loop_scope="function"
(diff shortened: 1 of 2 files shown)


jowilf:
starlette-admin>=1.0.0 support defaults for all backends

PR #781, opened 2026-06-29
Allow filtering on enum ListField

snow-born:
## Related
– Builds on: https://github.com/jowilf/starlette-admin/pull/746

## Summary
This PR adapts the multiple choice behaviour of `EnumField` by testing with `contains`, so that combinations of options from the enum are available for filtering.

### New Features
* Use datatables array operators `contains` and `without` to allow filtering of enum-type `ListField` columns.

### Modified Files
* `starlette_admin/statics/js/list.js` — Use the encapsulated `EnumField` within the `ListField`. Add new `enumArrayCondition` and share code between this and extant `enumCondition`.
* `starlette_admin/templates/list.html` — Bump `v=3` to `v=4` for `js/list.js` src path.

### Usage
Filter an enum `ListField` using the array operators `contains` and `without`.

image

### Testing
Manual verification of `ListField`, `EnumField` and enum `ListField`.

### DIFF
--- a/starlette_admin/statics/js/list.js
+++ b/starlette_admin/statics/js/list.js
@@ -29,7 +29,7 @@ $(function () {
           .concat(fringe);
       else if (field.type === "ListField") {
         // To reduce complexity, List of CollectionField will render as json
-        if (field.field.type == "CollectionField") {
+        if (field.field.type === "CollectionField" || field.field.type === "EnumField") {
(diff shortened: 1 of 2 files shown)


jowilf:
Sorry for not being able to check this on time, Datatable has been dropped and the new filter system already supports multiple choice for EnumField

PR #795, opened 2026-07-12
Validate next redirect target is same-origin

KezoSec:
Adds a same-origin check on the `next` query param used for post-login redirects, falling back to the admin index if it's off-origin.

Tests: confirms the legitimate deep-link → login → return flow still works, and confirms off-origin values are rejected.

No API changes. All existing tests pass (102 passed, 2 pre-existing unrelated skips).


jowilf (approved):
Thanks @KezoSec

